PR #12, opened 2017-08-16
updating react and react-dom to 15.6.1

ned2:
react-dom 15.4.2 seems to have disappeared from unpkg, so have upgraded external and local to 15.6.1

not sure if this is actually helpful but having got it all working on my dev environment with 15.6.1, figured I'd submit a PR


ned2:
it's about now I guess I should probably work out how to run the tests...

ned2:
Oh I see that the tests have v15.4.2 hardcoded... I guess that's the likely culprit

chriddyp:
Thanks @ned2 ! It seems like this issue was temporary from unpkg: https://github.com/unpkg/unpkg-website/issues/54 and that it is resolved. I don't think that changing the version will help us in the future here.
I've seen 2 issues since launch with Unpkg so I might consider loading these resources from a different CDN.

PR #21, opened 2017-09-15
Avoid excessive updates

chriddyp:
Fixes issue brought up in https://community.plot.ly/t/callback-is-fired-excessively-in-multi-app-setup-once-for-each-input-instead-of-once-for-each-callback/5789/4


chriddyp:
I have a preliminary fix available in `v0.10.0-rc1`. Try it out with:
```
pip install dash-renderer==0.10.0-rc1
```

chriddyp:
Just manually tested this on the [`dash-docs`](https://github.com/plotly/dash-docs) repo. With this patch, loading http://localhost:8000/dash/getting-started-part-2 results in 47% less requests (25 -> 17)! 🐎

wtluke:
No luck yet for me @chriddyp. LMK if you want to see my script to help debug (or decide it's a different issue)

chriddyp:
> LMK if you want to see my script to help debug (or decide it's a different issue)

@wtluke - Yeah, that would be helpful. Thanks!

wtluke:
Emailed 👍

chriddyp:
@wtluke - It looks like the infinite looping was a separate issue that was caused by having 2 `dcc.Location` properties with the same ID (`dcc.Location(id='url')`). (One in `runner.py`, the other in `beehive.py`). Removing the duplicate `dcc.Location` (in `beehive.py`) fixed the infinite loop for me ✅ 

Applying this patch to your example seemed to reduce the number of times one of your callbacks was being called from 16 times to just 2 times 🚀 !

wtluke:
Wow. Wow. Thank you, Chris. And this fix is insane - it has transformed the app into something unbelievably fast! Has caused me a lot of pain (trying to work out ways of avoiding 16 callbacks) - I never realised your solution here (calling it two times, per output) would have been possible. Wish I'd spoken up earlier. Absolutely superb - can't thank you enough. 

Onwards!

PR #26, opened 2017-10-17
allow any property to be a dash component, not just `children`

chriddyp:
pass in a connected `render` prop to every component and crawl _every_ prop since we don’t know where a component might be

related to https://twitter.com/chriddyp/status/920146214062043136


chriddyp:
closing in favor of https://github.com/plotly/dash-renderer/pull/32

PR #22, opened 2017-09-27
match response order with request order

chriddyp:
this PR will fix the issues found in https://github.com/plotly/dash/issues/133#issuecomment-332299141

PR #36, opened 2018-01-08
35 remove redux logger

### LINKED ISSUE #35: Remove redux-logger from production build
### Description
A large application floods the console with logging statements due to https://github.com/plotly/dash-renderer/blob/master/src/store.js#L18. There is already a TODO here to remove the logger. 

### Implementation discussion
It seems like this cannot be handled with webpack by a Dash user (e.g., use a `DEBUG` or production flag) because the `dash-renderer` bundle is included separately. Should the logger be removed entirely  for the `dash-renderer` production bundle? It is helpful for debugging during development.

mjclawar:
This PR uses the `process.env.NODE_ENV` to avoid adding logging to the redux middleware in the production build.
Closes #35

#### Other changes
- Adds `package-lock.json`
- Adds `.idea/` to gitignore for IntelliJ editors
- Bumps version & adds entry to CHANGELOG
- Adds small docstring to the `initializeStore` function.


chriddyp:
Looks good to me, thanks @mjclawar ! 

It looks like tests aren't running, I suspect because this is a forked branch. I'll update those settings.

chriddyp:
OK @mjclawar , I believe that forks should now trigger tests. Could you try pushing something up again?

mjclawar:
@chriddyp Tests all pass on my local machine -- any ideas why that one failed?

chriddyp:
Hm, I suspect that it's intermittent. I'll re-run the tests

chriddyp:
```
        input1.send_keys('hello world')
```
is supposed to fire the callback `len('hello world')` times: one for each character. However, https://github.com/plotly/dash-renderer/pull/22/ introduced some behaviour. From the changelog: "This fix should also improve performance when many updates happen at once as outdated requests will get dropped instead of updating the UI.". In this case, I suspect that some requests are getting processed out of order, requests are getting dropped, and the callback is fired 11 times instead of 12.

We should probably update the test with something like:
```
for char in 'hello world':
    input_1.send_keys(char)
    time.sleep(0.25)
```
or something more solid like
```python
text = 'hello world'
for (i, char) in enumerate(text):
    input_1.send_keys(char)
    wait_for_text_to_equal('output', text[:i])
```
Anyway, we don't need to fix that in this PR. I've logged the issue in https://github.com/plotly/dash-renderer/issues/38

mjclawar:
@chriddyp do you need anything else from me to merge this PR in and update? Thanks!

chriddyp:
@mjclawar Thanks for checking in! No I don't, just need to re-run CI a couple times until everything pases. Circle passes but it looks like Percy had a gateway timeout on the last run.

mjclawar:
@chriddyp Do you mind publishing 0.11.2 to npm when you get a chance? Thanks! The `__init__.py` will also need the [version](https://github.com/plotly/dash-renderer/blob/master/dash_renderer/__init__.py#L11) bumped to `0.11.2` to work I think.

chriddyp:
@mjclawar Thanks for reminder! I just published to npm and pypi

PR #45, opened 2018-02-11
Add optional React 16 support

mjclawar:
### Description
This adds optional React 16 support by allowing the developer to modify the React version served using, e.g.:  
```python
import dash_renderer
dash_renderer.REACT_VERSION = '16.2.0'
```

The default `REACT_VERSION` remains 15.4.2

### Other changes
- Updated .eslintignore to ignore all minified react bundles (versus just 15.4.2)
- Bumped version to 0.12.0


utyf:
The mechanism of choosing version is not clear. 
When you execute `import dash_renderer` file `dash_renderer/__init__.py` is executed and version is set to '15.4.2'. 
If you set `REACT_VERSION` to '16.2.0' later on the next line, this has no effect.

mjclawar:
@utyf You're totally right. I just realized this last night. I'm going to close this PR for now. There's another (ugly) monkeypatch example in this PR: https://github.com/plotly/dash-core-components/pull/160

utyf:
maybe we need to do something like this:
```python
def _set_react_version(react_version):
    assert react_version in _REACT_VERSION_TYPES
    # Dash renderer's dependencies get loaded in a special order by the server:
    # React bundles first, the renderer bundle at the very end.
    setattr(_this_module, '_js_dist_dependencies', [{
        'external_url': _REACT_VERSION_TO_URLS[react_version]['external_url'],
        'relative_package_path': _REACT_VERSION_TO_URLS[react_version]['relative_package_path'],
        'namespace': 'dash_renderer'
    }])

    setattr(_this_module, '_js_dist', [{
        'relative_package_path': 'bundle.js',
        "external_url": (
            'https://unpkg.com/dash-renderer@{}'
            '/dash_renderer/bundle.js'
        ).format(__version__),
        'namespace': 'dash_renderer'
    }])

_set_react_version('15.4.2')
```
and then, in the code:

```python
import dash_renderer
dash_renderer._set_react_version('16.2.0')
```

mjclawar:
That works. Adding and reopening with that. From the updated CHANGELOG:

> Allows user to now choose between '15.4.2' and '16.2.0' for React versions
> ```python
> import dash_renderer
> 
> # Set the react version before setting up the Dash application
> dash_renderer._set_react_version('16.2.0')
> 
> app = dash.Dash(...)
> ```

All previous tests still pass. Added another test to make sure that this actually updates the React version for a `dash.Dash` application initialized *after* `dash_renderer._set_react_version` has been called

Thanks, @utyf !

chriddyp:
This looks good to me! Thank you very much for adding the test. Anything else you would like to add @mjclawar before I merge?

mjclawar:
@chriddyp no I think this covers everything as long as you are OK with the `dash_renderer._set_react_version` method! 

Probably also will want to bump the default minimum React version to 15.6 soon to at least warn other dash component authors about `PropTypes` before ideally moving entirely to 16. Not sure if you want a more rigorous roadmap about updating `dash-renderer` React versions.

mjclawar:
@chriddyp anything else you need to see on this?

utyf:
Any news on the topic?

mjclawar:
@chriddyp Checking in to see if you want anything else here to accept the PR. 

We're looking forward to using React 16 for all our Dash applications!

chriddyp:
Sorry for the delay on this one! This looks good to me, I'll merge now and press a new release 🍻

chriddyp:
just published: `pip install dash-renderer==0.12.0`

PR #48, opened 2018-03-29
adapt MANIFEST.in to include all versions of React

utyf:
React 16 is not included in the package


mjclawar:
Thanks, @utyf, for catching this. I opened up another PR (https://github.com/plotly/dash-renderer/pull/50) before realizing this, and made #49 to track the issue.

chriddyp:
closed in #50

PR #52, opened 2018-04-04
Test to make sure data-* and aria-* actually render

rmarren1:


tests/test_render.py
```diff
@@ -1813,12 +1819,17 @@ def test_update_react_version(self):
         # Create a dummy component with no props
         # (dash-html-components may not support tested React version)
         class TestComponent(Component):
-            def __init__(self, _namespace):
+            def __init__(self, _namespace, _valid_wildcard_attributes, _type):
                 self._prop_names = []
                 self._namespace = _namespace
+                self._type = _type
```
rmarren1: This value not being present was causing an error in the `test_update_react_version` test.

PR #54, opened 2018-04-30
fix update behaviour for multiple properties

chriddyp:
currently, if a component updates two properties in a single `setProps` call and a callback is listening to both of those properties, the callback will be called twice instead of once.

this pr fixes that behaviour.

updating multiple properties at once is uncommon but was recently discovered when adding `n_clicks` and `n_clicks_timestamp` to the `dash_html_components` library: https://github.com/plotly/dash-html-components/pull/45

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -210,7 +210,17 @@ export function notifyObservers(payload) {
                     return;
                 }
                 InputGraph.dependenciesOf(node).forEach(outputId => {
-                    outputObservers.push(outputId);
+                    /*
+                     * Multiple input properties that update the same
+                     * output can change at once.
+                     * For example, `n_clicks` and `n_clicks_previous`
+                     * on a button component.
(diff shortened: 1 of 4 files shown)


chriddyp:
Could someone from @plotly/dash please review?

src/actions/index.js
```diff
@@ -210,7 +210,17 @@ export function notifyObservers(payload) {
...
+                    /*
+                     * Multiple input properties that update the same
+                     * output can change at once.
+                     * For example, `n_clicks` and `n_clicks_previous`
+                     * on a button component.
+                     * We only need to update the output once for this
+                     * update, so keep outputObservers unique.
+                     */
+                    if (!contains(outputId, outputObservers)) {
+                        outputObservers.push(outputId);
```
bpostlethwaite: Perhaps in a future iteration outputObservers could become a Set or similar data structure that enforces uniqueness.

PR #55, opened 2018-05-04
Added a CSS class to error messages.

pope1ni:
Added a CSS class to error messages output by dash on load to allow them to be styled.

A workaround for this is to use a (fragile) selector path to match the element, e.g.

```css
#react-entry-point > div > div:nth-child(2):not([id]):not([class])
```


chriddyp:
Makes sense, thanks!

Issue #57, opened 2018-06-21
When callback exceptions are supressed, if an `Input` or `State` references an ID that doesn't exist at callback trigger, then an front-end exception is thrown


chriddyp:

rmarren1:
More details:

when `app.config['suppress_callback_exceptions'] = False`, dash validates all callbacks when the application is launched and throws a `NonExistantIdException` [here](https://github.com/plotly/dash/blob/f309cc2b8a45644004db98ffcaf0d9a72870c27d/dash/dash.py#L381) if the _initial_ state of the layout does not include an `id` used in either the `Input`, `Output`, or  `State` of a callback.

E.g.,

```
import dash
import dash_html_components as html
import dash_core_components as dcc
from dash.dependencies import Input, Output, State

app = dash.Dash()
app.layout = html.Div([
    html.Button(id='button', children='click'),
    html.P(id='output')
])

@app.callback(Output('output', 'children'),
              [Input('button', 'n_clicks')],
              [State('buttoN', 'n_clicks')])
def put_n_clicks(n_clicks, n_clicks_state):
    if n_clicks:
        return n_clicks
    return 0

app.run_server(debug=True)

```
-->
```
dash.exceptions.NonExistantIdException: 
Attempting to assign a callback to the
component with the id "buttoN" but no
components with id "buttoN" exist in the
app's layout.


Here is a list of IDs in layout:
['button', 'output']
```
This exception would be raised if a user includes a callback using an `id` of a component that does not exist in the _initial_ state, but is dynamically generated by other callbacks.

For example, an exception is raised in this example:
```
import dash
import dash_html_components as html
import dash_core_components as dcc
from dash.dependencies import Input, Output, State

app = dash.Dash()
app.layout = html.Div([
    html.Button(id='button', children='click'),
    html.Div(id='output-div-holder'),
    html.Div(id='output-paragraph')
])

@app.callback(Output('output-div-holder', 'children'),
              [Input('button', 'n_clicks')])
def put_n_clicks(n_clicks):
    return html.Div(children=n_clicks, id='output-div') if n_clicks else 'None'

@app.callback(Output('output-paragraph', 'children'),
              [Input('output-div', 'children')])
def put_child(child):
    return child

```
because the second callback has an `Input` for the id `'output-div'`, but that component does not exist in the initial state (it is generated once the user clicks the button).

[...]

PR #60, opened 2018-07-16
Check that id exists before triggering callback.

rmarren1:
Fixes https://github.com/plotly/dash-renderer/issues/57

Before creating payloads for input and state, make sure the id of each input and state actually exists in the current graph
image

image


src/actions/index.js
```diff
@@ -417,8 +417,21 @@ function updateOutput(
             dependency.output.property === outputProp
         )
     );
+    const validKeys = Object.keys(paths);
```
chriddyp: By convention, we use [`ramda`](https://ramdajs.com/docs/) as our "standard library" for all things JavaScript, even when a JavaScript method exists. So here, this would be `R.keys(paths)`

src/actions/index.js
```diff
@@ -417,8 +417,21 @@ function updateOutput(
...
     if (inputs.length > 0) {
         payload.inputs = inputs.map(inputObject => {
+            // Make sure the component id exists in the layout
+            if (!validKeys.includes(inputObject.id)) {
```
chriddyp: and in Ramda, this would be `!R.contains(inputObject.id, validKeys)`

chriddyp:
Really good stuff, nice work getting into `dash-renderer`, I believe you are one of the first to do so!

Would you mind also updating the:
- [ ] `CHANGELOG.md`
- [ ]  `package.json` version number
- [ ] and the `version.py` version number

That way, we can release as soon as we merge.

💃 once those are updated and the `ramda` syntax is updated

Issue #29, opened 2017-11-22
`true`/`false` as `children` in layout crashes app


chriddyp:
see https://community.plot.ly/t/bug-in-tables-boolean-values-fail-to-render/6872

chriddyp:
Also see https://github.com/plotly/dash/issues/264

rmarren1:
To prevent this crash, we would just need to add 'Boolean' to the arrays [here](https://github.com/plotly/dash-renderer/blob/70bb93e8f0151ddc7d0810f3df271036a069360a/src/TreeContainer.js#L24) and [here](https://github.com/plotly/dash-renderer/blob/70bb93e8f0151ddc7d0810f3df271036a069360a/src/TreeContainer.js#L42).
![image](https://user-images.githubusercontent.com/10272301/42791374-531ba684-893e-11e8-84de-b7955b58adea.png)

This results in a Javascript boolean being passed to the react `render` function, so nothing is actually rendered on the screen (left in picture) but the boolean value still exists in the react dom (right in picture).

From my own experience, a common use case for setting `children` as boolean is when many callbacks need to run an *expensive* boolean function, you can first store the result of the function in the front end. This fix works for that use case, and if a user wants to actually render the boolean they can cast it to a string first.

Does @plotly/dash think this is a good idea? Or should we actually render the strings 'true' or 'false' to the page. Users can just set `style={'display': 'none'}` when they do not want to render.

valentijnnieman:
I don't really see a reason why we would render a boolean as a string - like you said, if you'd want that, you can easily cast it to a string first. I also don't see a reason why *not* to allow booleans to be passed, and if there's a use-case for it, I'd say go for it!

rmarren1:
Fixed in #62

PR #62, opened 2018-07-19
Don't throw front end exception when rendering boolean

rmarren1:
Small fix to https://github.com/plotly/dash-renderer/issues/29

When a user attempts to render a boolean value, nothing will be visible on screen (like when None is rendered). Users can cast to string for this functionality. This seemed like the correct default (as opposed to defaulting to rendering a string and requiring users to supply `style={'display': 'none'}`).


chriddyp:
💃  Much better than just failing!

Nice work getting in there 🙌

PR #69, opened 2018-08-21
Add 'namespace' and 'type' props to _dash_update_component payload.

rmarren1:


chriddyp:
This looks good to me! Let's keep it open until the rest of https://github.com/plotly/dash/pull/340 is worked through

alexcjohnson:
This little PR is required for https://github.com/plotly/dash/pull/452, which is the successor to https://github.com/plotly/dash/pull/340. I'm going to close it though, since it's out of date and will get even more so once multi-outputs #91 are merged, and it will be easy enough to recreate when we're ready for it.

PR #72, opened 2018-08-31
Update webpack

T4rk1n:
Update webpack to version 4. Allow debugging with the inline source map of `build:dev`.


valentijnnieman:
Nice! Looks good to me 💃

PR #73, opened 2018-08-31
Hot reload

T4rk1n:
**Available with the following version:**
- `dash==0.27.0rc7`
- `dash-renderer==0.14.0rc6`

Add hot-reload capability on project file changes by periodically requesting a hash and checking it's the same hash as the initial one.

Adds:
- Reloader component
- reloadRequest api thunk.

Need plotly/dash#362


T4rk1n:
@plotly/dash  Ready for review!

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export default connect(
+    state => ({
+        config: state.config,
+        reloadHash: state.reloadHash
+    }),
+    dispatch => ({dispatch})
+)(Reloader);
+
```
rmarren1: newline

src/actions/api.js
```diff
@@ -106,3 +106,11 @@ export function login(oauth_token) {
...
     );
 }
+
+export function getReloadHash() {
+    return apiThunk(
+        '_reload-hash',
+        'GET',
+        'reloadRequest'
+    )
+}
```
rmarren1: newline

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    componentDidMount() {
+        const { dispatch } = this.props;
+        const { disabled, interval } = this.state;
+        if (!disabled && !this._intervalId) {
+            this._intervalId = setInterval(() => {
+                if (!this.state.reloading) {
+                    dispatch(getReloadHash());
```
rmarren1: Should `state.reloading` be set to `true` here? I do not see anywhere that `state.reloading` is changed, it seems to always be `false`.

src/reducers/reducer.js
```diff
@@ -89,4 +90,14 @@ function recordHistory(reducer) {
     }
 }
 
-export default recordHistory(reducer);
+function rootReducer(reducer) {
+    return function(state, action) {
+        if (action.type === 'RELOAD') {
+            const {history} = state;
```
rmarren1: I there extra stuff in `state` that is stripped when you do this? I don't see what these two lines do, I don't think `state` is changed here

src/components/core/Reloader.react.js
T4rk1n: I think it's a leftover from something I tried, I will remove thanks.

src/reducers/reducer.js
T4rk1n: Everything but the history is stripped down, it triggers a soft reload.

rmarren1: Oh okay so everything else gets set to default except the history, looks good

valentijnnieman: Hmm, can you not combine this reducer into the `recordHistory` function that's already there? I'm not sure I understand why it is a separate reducer.

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,85 @@
...
+    componentDidUpdate() {
+        const {reloadHash, dispatch} = this.props;
+        if (reloadHash.status === 200) {
+            if (this.state.hash === null) {
+                this.setState({hash: reloadHash.content.reloadHash});
+                return;
+            }
+            if (reloadHash.content.reloadHash !== this.state.hash) {
+                // eslint-disable-next-line no-undef
+                window.clearInterval(this._intervalId);
```
valentijnnieman: So the `no-undef` rule here should not trigger if it's sure that `this._intervalId` is set, right? Maybe it's better to write a check here that makes sure that it's not `null`, instead of disabling the rule.

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,85 @@
...
+                if (reloadHash.content.hard) {
+                    // Assets file have changed, need to reload them.
+                    // eslint-disable-next-line no-undef
+                    window.top.location.reload();
```
valentijnnieman: Hmm, why is the `no-undef` rule triggering here?

src/reducers/reducer.js
```diff
@@ -20,6 +20,7 @@ const reducer = combineReducers({
     dependenciesRequest: API.dependenciesRequest,
     layoutRequest: API.layoutRequest,
     loginRequest: API.loginRequest,
+    reloadHash: API.reloadRequest,
```
valentijnnieman: Any reason why it's named `reloadHash` here and `reloadRequest` on `API.reloadRequest`?

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,85 @@
...
+                hash: null,
+                interval,
+                disabled: false
+            }
+        } else {
+            this.state = {
+                disabled: true
+            }
+        }
+        this._intervalId = null;
```
valentijnnieman: Can `_intervalId` not be in `state` as well?

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,85 @@
...
+class Reloader extends React.Component {
+    constructor(props) {
+        super(props);
+        if (props.config.hot_reload) {
+            const { interval } = props.config.hot_reload;
+            this.state = {
+                hash: null,
+                interval,
+                disabled: false
+            }
```
valentijnnieman: What if this state is also handled in a reducer? Would there be value in having this state in the store? Perhaps saving the `hash` in the store for example would make it easier to at some point in the future to enable the store to be saved client-side, so that it can be rehydrated upon refreshing.

src/components/core/Reloader.react.js
T4rk1n: The `no-undef` is for `window`, it highlight as error in pycharm.

src/reducers/reducer.js
T4rk1n: Good catch, I refactored `reloadHash` to `reloadRequest` to conform to the other API, I guess that one didn't follow up.

src/components/core/Reloader.react.js
T4rk1n: The hash is null at first, it is set in the state from the server on the first hash request so I don't think it needs to be rehydrated.

src/components/core/Reloader.react.js
valentijnnieman: Ah ok! Hmm maybe that rule can be turned off globally then, if it's always going to trigger on `window`?

valentijnnieman: Last question: isn't it maybe better to put the `rootReducer` in a separate file, and call it something else, like `reloadReducer` or whatever?

T4rk1n: > maybe that rule can be turned off globally then

I think we can put `window` in a global exclude or something. Anyway it underline as error in IDE, I just hit `alt-enter` and select a fix, doesn't take more than a second.

> Last question: isn't it maybe better to put the rootReducer in a separate file

It's only a couple lines and is the last reducer exported, keeping it in `reducer` make more sense to me.

src/reducers/reducer.js
rmarren1: I think it should be separate from `recordHistory`. This reducer doesn't really change the history, it just removes everything else. It could have a different name though for sake of clarity, maybe like `preserveHistoryOnReload`

src/components/core/Reloader.react.js
```diff
@@ -0,0 +1,155 @@
...
+                            const nodesToDisable = [];
+
+                            // Search for the old file by xpath.
+                            const it = document.evaluate(
+                                `//link[contains(@href, "${a.url}")]`,
+                                this._head
+                            );
+                            let node = it.iterateNext();
+
+                            while (node) {
```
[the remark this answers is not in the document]
T4rk1n: It's a proper iterator but it doesn't work, if there's no element found (asset file added) it will fail to make an array with the iterator.

Issue #65, opened 2018-08-06
Request Hooks


chriddyp:
The API requests that Dash's front-end makes should be configurable. In order to make them configurable, we will need to do the following things:

- We will need to make the Dash Front-end (`dash-renderer`) a standalone library that can be initialized and invoked from other libraries
- We will need to extend the `dash-renderer` to take a set of config input parameters
- Some of these input parameters will be functions or promises that we will call before certain actions
- For this issue, we will start by a `request_pre` and `request_post` parameter that will be called before and after HTTP requests with the entire request and response objects. This will enable the developer to transform the data before and after network requests.
Example:
```
dash_renderer({
    request_pre: requestPayload => {
        return modifyRequest(requestPayload);
    },
    request_post: (requestPayload, responsePayload) => {
        return modifyResponse(requestPayload, responsePayload)
    }
})
```

So, the hooks are only the network requests. In the code, this would fit in here: https://github.com/plotly/dash-renderer/blob/70bb93e8f0151ddc7d0810f3df271036a069360a/src/actions/index.js#L447-L455

Note that in order for users to supply their own arguments, I believe they would need to provide their own instance of `dash_renderer` by overriding the index file (as discussed here:  https://github.com/plotly/dash/issues/265)

alexcjohnson:
This was released in #75

nirvana-msu:
It seems that certain types of responses do not trigger `request_post`. Namely:

1. On initial load (i.e. not in response to user interaction) only `request_pre` is fired but not `request_post`.
2. There is no hook for failed responses.

I am currently using [fetch-intercept](https://github.com/werk85/fetch-intercept) to provide custom loading indicators and I was hoping I could switch to custom Dash hooks but it's not possible due to issues describved above. `fetch-intercept`, on the other hand,, works perfectly as it provides me with `responseError` hook and it always fires (including on initial load).

Separately, it'd be nice if the hooks provided access to the actual Fetch request / response objects to inspect or modify.. I'll have to stick with `fetch-intercept` for now.

alexcjohnson:
Please make an issue for the problems you've encountered at https://github.com/plotly/dash/issues

This repo is obsolete, the renderer has been merged into the main dash repo.

PR #75, opened 2018-09-05
Custom hooks - March 1

valentijnnieman:
This adds customizable hooks before and after the fetch to `dash-update-component`, and makes the `dash-renderer` a standalone class called `DashRenderer` that can be called with an optional object holding `request_pre` and `request_post` functions, a la #65. Here's a simple example: 
```
app.index_string = '''
<!DOCTYPE html>
<html>
    <head>
        {%metas%}
        <title>{%title%}</title>
        {%favicon%}
        {%css%}
    </head>
    <body>
        <div>Testing custom DashRenderer</div>
        {%app_entry%}
        <footer>
            {%config%}
            {%scripts%}
            <script id="_dash-renderer" type"application/json">
                const renderer = new DashRenderer({
                    request_pre: (req) => {
                        console.log('req:', req);
                        var output = document.getElementById('output-pre')
                        if(output) {
                            output.innerHTML = 'request_pre was fired and returned: <br/>req: ' + req + '<br/><br/>';
                        }
                    },
                    request_post: (req, res) => {
                        console.log('res:', res);
                        var output = document.getElementById('output-post')
                        if(output) {
                            output.innerHTML = 'request_post was fired and returned: <br/>res: ' + JSON.stringify(res) + '<br/><br/>';
                        }
                    }
                })
            </script>
        </footer>
        <div>With request hooks</div>
    </body>
</html>
'''

app.layout = html.Div([
    dcc.Input(
        id='input',
        value='initial value'
    ),
    html.Div(
        html.Div([
            html.Div(id='output-1'),
            html.Div(id='output-pre'),
            html.Div(id='output-post')
        ])
    )
])

@app.callback(Output('output-1', 'children'), [Input('input', 'value')])
def update_output(value):
    return value
```

It uses the new customizable `index_string` to setup a custom `DashRenderer` instance.

This depends on `dash` having support for a standalone `dash-renderer`, so [this PR](https://github.com/plotly/dash/pull/367) should be merged along side of this one.


src/DashRenderer.js
```diff
@@ -0,0 +1,29 @@
...
+'use strict';
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import AppProvider from './AppProvider.react';
+
+class DashRenderer {
+    constructor(hooks = { request_pre: null, request_post: null}) {
+        // do hooksstuff here, like custom hooks
```
chriddyp: Is there anything else that needs to be done in here or can we ⚡️ this comment?

src/DashRenderer.js
```diff
@@ -0,0 +1,29 @@
...
+        );
+    }
+}
+
+DashRenderer.propTypes = {
+    hooks: PropTypes.shape({
+        request_pre: PropTypes.func,
+        request_post: PropTypes.func,
+    })
+}
```
chriddyp: These aren't actually props right? We're referring to them in the constructor rather than in `this.props` and `DashRenderer` doesn't extend `Component`

src/actions/index.js
```diff
@@ -469,6 +471,19 @@ function updateOutput(
         });
     }
 
+    if(hooks.request_pre !== null) {
+        if(typeof hooks.request_pre === 'function') {
+            hooks.request_pre(JSON.stringify(payload));
```
chriddyp: I think that one of the points of these hooks is the ability to modify the `payload`. So, we shouldn't make a copy of this, we should leave it up to the developer to modify it at will

src/actions/index.js
```diff
@@ -469,6 +471,19 @@ function updateOutput(
         });
     }
 
+    if(hooks.request_pre !== null) {
+        if(typeof hooks.request_pre === 'function') {
```
chriddyp: For consistency, I try to use `ramda` throughout this library. So, this would be `if (R.type(hooks.request_pre) === 'function')

src/actions/index.js
```diff
@@ -732,8 +763,8 @@ function updateOutput(
             }
 
         });
+    }).then(() => {
```
chriddyp: ⚡️ this?

src/APIController.react.js
```diff
@@ -49,6 +51,10 @@ class UnconnectedContainer extends Component {
             }
         }
 
+        if(hooks.request_pre !== null || hooks.request_post !== null) {
+            dispatch(setHooks(hooks));
```
chriddyp: I believe that `dispatch` is asynchronous, so if we want to hydrate in this function (which makes sense, it's where we hydrate our API calls and our derived data), then we need to add a check below that `setHooks` has finished before we call `dispatch(hydrateInitialOutputs());`

chriddyp: That is, in here: https://github.com/plotly/dash-renderer/pull/75/files#diff-1b2b6e62305d1ce2aaa047bb80a6f529R69

chriddyp: The `dash-renderer` initialization is handled entirely in this logic:
1. The `layout` and `dependencies` API calls are made `dispatch(getLayout())`
2. The `paths` (`dispatch(computePaths({...}))` and `graphs` (`dispatch(computeGraphs(...))` stores are computed
3. Once that data is computed, we "hydrate" the initial outputs, that is we make the API calls to fire the callbacks for all of the outputs on the page `dispatch(hydrateInitialOutputs())`
4. Once everything is "hydrated", we render the app: ` else if (appLifecycle === APP_STATES('HYDRATED')) {`

These `dispatch` calls are asynchronous. So, when a `dispatch(...)` call is made, some stuff gets recomputed, and the component gets re-renderered, which fires this `initialization` call. This `initialization` call is repeatedly called until all of the stores are computed and the order in which stores are computed is described through the `if` statements. That is, this block:
```
        if (isEmpty(dependenciesRequest)) {
            dispatch(getDependencies());
        } else if (dependenciesRequest.status === 200 && isEmpty(graphs)) {
            dispatch(computeGraphs(dependenciesRequest.content));
}
```

is implicitly saying: first make the API call for the dependencies, _then_ compute the graphs based off the dependencies. The "_then_" is implicitly understood because of the lifecycle of the app: once the API call is finished, the component is _re-rendered_, firing this `initialization` function again (because `initialization` is fired in `componentWillReceiveProps`).

> As a side note, the alternative way to write this would be through a gnarly promise chain: `dispatch(getLayout()).then(dispatch(setLayout()).then(dispatch(computePaths(...))...` etc. I found that it's easier to reason about the order in which actions depend on each other (and the failure cases) through this more declarative representation where this function is just repeatedly called.

So, in this PR, you're introducing a new store. It's `null` by default and, since it affects the API calls, we need to make sure that it is computed before we start hydarting. You've set up the trigger to compute the hook, but you haven't set up any logic to prevent the hydration to happen before the hooks are finished. So, you need to describe that in https://github.com/plotly/dash-renderer/blob/bd3c8b38a23c4c953637d0518db00298f0f3d0c3/src/APIController.react.js#L58-L72, by adding another `if` statement that says "make sure that the hooks are set up before we start making the api calls". How you set up that if statement is up to you. In the other cases, we initialize the store to be `{}` or `null` and then we simply check if the store is _not_ `{}` or `null` (if it's not `{}` or `null`, then we know that it was computed). A similar check would probably work here too. Or, you could initialize the store to be some constent like `constants.EMPTY = 'EMPTY'` and then check if it's not `EMPTY`. But I think that `null` would work too.

I wouldn't use `status` or `200` here because those are used just for the API requests.

valentijnnieman:
@chriddyp Thanks for your detailed explanation! Had to write some logic for the case of hooks not being used, hope I got it right.

chriddyp:
Another solution might be to place it into the store when we create it, that is here: https://github.com/plotly/dash-renderer/blob/bd3c8b38a23c4c953637d0518db00298f0f3d0c3/src/AppProvider.react.js#L7-L11. Perhaps `store.initializationConfig = {hooks: hooks}`. Or maybe there is another way to do this with the redux reducer initialization sequence. However, that seems a little less straightforward / standard than just wiring everything through.

valentijnnieman:
@bpostlethwaite Yeah, isn't the `context` api part of React 16 though? 

I'm also wondering: if I thread it all the way down, won't it fire for each component wrapped in a `<NotifyObservers/>` component? Will look into this more when I have a chance, @chriddyp sounds like you're right about it not having to be in the store though. Thanks!

bpostlethwaite:
Context has been part of the React API for a long time - it's recently been elevated out of "don't use this" status. But still it's probably best not to use it for the fact that it can make code harder to test / reason about.

valentijnnieman:
> where we're firing the actions: ➡️ TreeContainer ➡️ NotifyObservers and then wire it in as arguments in the dispatch(notifyObservers()) API calls that the NotifyObservers component makes.

Won't that pass the `hooks` down to all components though? Feels like it shouldn't have to be passed down to each individual one, hence why I put it in the store. 
What about passing the hooks to `hydrateInitialOutputs` and set them in the store from there, so that they're available each time `updateOutput` fires?

valentijnnieman:
@chriddyp Actually, thinking about it again, I don't think the `dispatch(setHooks)` call is async at all. It's not using any sort of thunk middleware, so I believe in this case the hooks will be set before anything else happens, so the previous version (without checking if the hooks are set) should be good. 

Here's a [stack overflow question about it](https://stackoverflow.com/questions/43276291/is-store-dispatch-in-redux-synchronous-or-asynchronous)

chriddyp:
> Won't that pass the hooks down to all components though? 

No it won't, it will only pass it down to the `NotifyObservers` component, which the user doesn't have access to. It won't be in the actual `props` of the component.

> I'm also wondering: if I thread it all the way down, won't it fire for each component wrapped in a <NotifyObservers/> component?

I'm not sure what you mean by "fire". It won't cause any extra renders though, because it will just be integrated into the initial props of all of the parents.

> so I believe in this case the hooks will be set before anything else happens

Can you verify this yourself by putting in debugger calls? 

> What about passing the hooks to hydrateInitialOutputs and set them in the store from there

The main thing that I'm concerned about is that the order of operations needs to be very clear. If we have this done in `APIController`, the dependency of setting hooks needing to be finished before `updateOutput` calls is implicit rather than explicit.

valentijnnieman:
@chriddyp I meant that, because the `NotifyObserver` component wraps all the components that are being rendered/passed from Dash to dash-renderer, if you push a `hooks` prop down, that would mean that every component's `NotifyObserver` wrapper would get that `hooks` prop and will fire (dispatch, sorry) the `setHooks` action for every component, while we just want it to fire once in the beginning to update the store with the hooks. Obviously the prop won't be available to users :-)

I'll revert an reverify if I'm correct about that action firing first - it makes sense though, whichever action comes first should be handled first.

valentijnnieman:
@chriddyp I'm not so sure if that's going to work - I'm trying to pass the `hooks` down to the `NotifyObservers` components from the `TreeContainer`, but because of that recursive `render` function that's also used as a `.map` (on line 55 in `TreeContainer.js`) it sometimes passes in an index . 
screen shot 2018-10-11 at 2 47 53 pm
I can write some logic etc but I feel that this is getting so convoluted, while it should be relatively simple, which is slightly frustrating, haha.

chriddyp:
> I'm trying to pass the hooks down to the NotifyObservers components from the TreeContainer, but because of that recursive render function that's also used as a .map

Hm yeah, that does seem confusing. OK, perhaps an initial dispatch would be easier. Could we find a place to put it higher up than in the tree than `APIController`? That way, we can keep the actions inside `APIController` only related to the API calls.

valentijnnieman:
@chriddyp Hmm it looks like you were right though, `hydrateInitialOutputs()` does fire before the hooks.
screen shot 2018-10-11 at 3 09 47 pm

.circleci/config.yml
```diff
@@ -45,8 +45,8 @@ jobs:
               - "node_modules"
 
       - run:
-          name: Npm lint
-          command:  npm run lint
+          name: Prettier
+          command:  npm run format:test
```
alexcjohnson: FYI @valentijnnieman I added this formatting commit. I noticed some linting issues and was surprised the tests didn't catch them... turns out
- `npm run lint` as currently defined is a fixer, not a tester, so it can't fail. Anyway we want to test `prettier` since it's stricter (is there actually a purpose to `eslint` if we're running `prettier` too?)
- `format` and `format:test` didn't inspect `.react.js` files - I guess `*.js` doesn't match `*.react.js`. I fixed that in `package.json`.

alexcjohnson: Oh I see... `prettier` only does formatting, not code quality checks, didn't realize that. I'll update so this runs both (and both as tests!)

valentijnnieman: `prettier` can test code formatting too, if that's what we want. I set that up in core components I believe

valentijnnieman: As here: https://github.com/plotly/dash-core-components/blob/83a3ea07fd8af9ee3ca57c8da8ed32483e96f87c/package.json#L25

alexcjohnson: Hmm, they seem to say otherwise https://prettier.io/docs/en/comparison.html

> Code-quality rules: eg no-unused-vars, no-extra-bind, no-implicit-globals, prefer-promise-reject-errors...
> 
> Prettier does nothing to help with those kind of rules. They are also the most important ones provided by linters as they are likely to catch real bugs with your code!

Anyway I set it to run both https://github.com/plotly/dash-renderer/pull/75/commits/6c309d714de747b324e109b5ccaa76ad2e9fdbb4

BTW it would be worth looking at this `.react.js` issue in the other repos...

Also in the main `dash` repo I also made all of these tests run to completion, but I didn't do that here, this setup will still quit on first failure. I prefer to get all my errors at once, but seems like I may be in the minority on that issue...

valentijnnieman: Yeah, I was talking about testing formatting - it'll throw an error if the code isn't formatted the way `prettier` wants it. I like it personally - that way I can actually use `prettier` and not worry about committing lots of noise regarding formatting changes. But I know others aren't so keen on it! Anyway, yeah, it doesn't replace linting or anything. 

Not sure what you mean about the tests - aren't these ones all being run to completion too?

src/actions/index.js
```diff
@@ -555,6 +575,20 @@ function updateOutput(
...
+            // Fire custom request_post hook if any
+            if (hooks.request_post !== null) {
+                if (type(hooks.request_post) === 'Function') {
+                    hooks.request_post(payload, data.response);
+                } else {
+                    /* eslint-disable no-console */
+                    // Throwing an Error or TypeError etc here will cause an infinite loop for some reason
+                    console.error(
+                        'The request_post hook provided was not of type function, preventing Dash from firing it. Please make sure the request_post hook is a function'
+                    );
```
alexcjohnson: Nonblocking, but probably would have been better to declare that these need to be functions in somebody's `propTypes` and avoid addressing this possibility explicitly.

tests/test_render.py
```diff
@@ -1898,6 +1898,79 @@ def update_output(*args):
...
+            </head>
+            <body>
+                <div>Testing custom DashRenderer</div>
+                {%app_entry%}
+                <footer>
+                    {%config%}
+                    {%scripts%}
+                    <script id="_dash-renderer" type"application/json">
+                        const renderer = new DashRenderer({
+                            request_pre: () => {
```
alexcjohnson: Can we use the arguments to these hooks in this test? Just to ensure that the right things are there and that they don't change unintentionally.

tests/test_render.py
```diff
@@ -1898,6 +1898,79 @@ def update_output(*args):
...
+
+        action = ActionChains(self.driver)
+        action.click(input1)
+        action = action.send_keys(Keys.BACKSPACE * len(initialValue))
+
+        action.send_keys('fire request hooks').perform()
+
+        self.wait_for_text_to_equal('#output-1', 'fire request hooks')
+        self.wait_for_text_to_equal('#output-pre', 'request_pre changed this text!')
+        self.wait_for_text_to_equal('#output-post', 'request_post changed this text!')
```
alexcjohnson: This feels like it could be flaky - like if `request_post` fires too soon after `request_pre` so we never see `'request_pre changed this text!'`. Perhaps add a short delay (half second?) in the callback to make sure this doesn't happen?

.circleci/config.yml
alexcjohnson: The way I did it in https://github.com/plotly/dash-renderer/pull/75/commits/6c309d714de747b324e109b5ccaa76ad2e9fdbb4, CI runs `lint:test`, and if that fails the whole test run stops. If it succeeds we go on to `format:test`, and if that fails the whole test run stops. Only if both of those pass do we even run the python tests.

`dash`, on the other hand, is currently [set up](https://github.com/plotly/dash/blob/master/test.sh) to run all of these no matter what, but still registering a failure if any of them failed. I prefer that, so I can go back and fix all the errors at once, but seems like others mostly prefer to get an answer more quickly.

tests/test_render.py
alexcjohnson: Oh sorry - missed that these are 2 different elements. This is fine as is.

tests/test_render.py
valentijnnieman: Yeah good call. I've added them and I'm doing an assertion on the output. However, if this output ever changes, this test will automatically fail. Are we ok with that?

tests/test_render.py
```diff
@@ -1918,17 +1918,25 @@ def test_request_hooks(self):
...
+                                var outputPayload = document.getElementById('output-pre-payload')
                                 if(output) {
                                     output.innerHTML = 'request_pre changed this text!';
                                 }
+                                if(outputPayload) {
+                                    outputPayload.innerHTML = JSON.stringify(payload);
+                                }
                             },
-                            request_post: () => {
+                            request_post: (payload) => {
```
alexcjohnson: `request_post` also receives the response - we should test something about at least the basic structure of this response.

src/DashRenderer.js
```diff
@@ -0,0 +1,34 @@
...
+    constructor(hooks) {
+        // render Dash Renderer upon initialising!
+        ReactDOM.render(
+            <AppProvider hooks={hooks} />,
+            document.getElementById('react-entry-point')
+        );
+    }
+}
+
+DashRenderer.propTypes = {
```
alexcjohnson: Oh, per @chriddyp 's earlier comment, `DashRenderer` isn't a react component so it doesn't have props. But `AppProvider` is a component, so this can go there.

Can we add a test that tries to send non-function hooks and show that it fails in an intelligible way?

tests/test_render.py
alexcjohnson: > if this output ever changes, this test will automatically fail. Are we ok with that?

In fact that's exactly why we want this in the tests - so we know it changed and we can let custom hooks users know they will need to update their code!

alexcjohnson:
@chriddyp pretty sure all your comments from last fall have been long since addressed - can I dismiss your review?

chriddyp:
> can I dismiss your review?

👍 💃 !

PR #81, opened 2018-09-20
Filter nil values before triggering initial layout callbacks.

T4rk1n:
This get rid of the initial None callbacks, during my investigation (plotly/dash#396), I found that the None callbacks were blocking more useful callbacks.

We would need to update all the docs that does this:

```python
@app(Output('output', 'children'), [Input('btn', 'n_clicks')])
def on_clicks(n_clicks):
    if n_clicks is None:
        raise PreventUpdate
```


chriddyp:
I'm hesitant to do this right now as it would be a breaking change. I think that I'd like to combine a set of breaking changes together as part of a greater 2.0 release that we could more easily document and inform our community about.

For example, if we changed this behaviour, then we could introduce the Ryan's "components with defaults" PRs (https://github.com/plotly/dash-core-components/pull/234) among other features.

Perhaps we should create a 2.0 master issue to document all of these breaking changes that we could / would like to do in the next version. Similar to plotly.js's 2.0 issue: https://github.com/plotly/plotly.js/issues/420

PR #89, opened 2018-10-11
add missing polyfill

chriddyp:
see context here https://github.com/plotly/dash-renderer/issues/87#issuecomment-429155275

for immediate review anyone from @plotly/dash


chriddyp:
Just tested IE10 on browserstack too
image

PR #91, opened 2018-10-25
[WIP] Add multi output callback support.

T4rk1n:
Needs plotly/dash#436

Tests are gonna fail until a rc version of dash has been released.

`pip install dash-renderer==0.19.0rc1` try with `dash==0.38.0rc1`


src/actions/index.js
```diff
@@ -560,180 +561,189 @@ function updateOutput(
...
-                             * And none of its inputs are generated in this
-                             * request
-                             */
-                            intersection(
-                                InputGraph.dependantsOf(idAndProp),
-                                keys(newProps)
-                            ).length === 0
-                        ) {
-                            outputIds.push(idAndProp);
-                            delete newProps[idAndProp];
```
chriddyp: It's pretty hard for me to read the changes in this file - A lot of code hasn't changed, it was just moved and indented. When this is ready for review, could we break up these changes into several commits, so that I can more easily determine which lines were changed vs just moved?

T4rk1n: That is prettier fault, it completely messes up when there's wrong indent mixed up.

bpostlethwaite: I can't tell if you are being sarcastic or not but in case you are not... prettier doesn't run git or push code to Github and it isn't responsible for making PRs pleasant to review. You are responsible for these things.

My usual workflow when developing on un-pretty things is to open the unpretty file, add a space, save... so that prettier runs. Then I commit that as "prettify index.js". Then subsequent commits make your actual logical commits. 

That way a reviewer can step through your PR commit by commit like a book! It's the developers responsibility to make sure reviewing is good experience :)

Also git rebase and rewrite a messy history so your PR has a nice sequence of commits that can be stepped through. Add longer commit descriptions where appropriate. 

Spend time making the life of your reviewers pleasant!

T4rk1n: Sorry, should have committed before running prettier. Thought it would only add missing `,`, but the comments had bad indent so it messed up.

src/actions/index.js
```diff
@@ -242,7 +242,15 @@ export function notifyObservers(payload) {
...
-            const outputComponentId = outputIdAndProp.split('.')[0];
+            let outputIds;
+            if (outputIdAndProp.startsWith('[')) {
+                outputIds = outputIdAndProp
+                    .slice(1, outputIdAndProp.length - 1)
+                    .split(':')
+                    .map(e => e.split('.')[0]);
+            } else {
+                outputIds = [outputIdAndProp.split('.')[0]];
+            }
```
T4rk1n: Create an array of output ids instead of a single output.

src/actions/index.js
```diff
@@ -316,9 +324,12 @@ export function notifyObservers(payload) {
              * of a controller change.
              * for example, perhaps the user has hidden one of the observers
              */
+
+            const {paths} = getState();
+
             if (
                 controllersInFutureQueue.length === 0 &&
-                has(outputComponentId, getState().paths) &&
+                any(e => has(e, paths))(outputIds) &&
```
T4rk1n: Modify the check to account for multi-output.

src/actions/index.js
```diff
@@ -364,8 +374,7 @@ export function notifyObservers(payload) {
 }
 
 function updateOutput(
-    outputComponentId,
-    outputProp,
+    outputIdAndProp,
```
T4rk1n: Modified the dash side handling of dispatch to only take the string key of the output.

src/actions/index.js
```diff
@@ -391,17 +400,15 @@ function updateOutput(
...
 
     if (event) {
         payload.event = event;
     }
 
     const {inputs, state} = dependenciesRequest.content.find(
-        dependency =>
-            dependency.output.id === outputComponentId &&
-            dependency.output.property === outputProp
+        dependency => dependency.output === outputIdAndProp
```
T4rk1n: Only check the output id since that is what we now pass in `_dash-dependencies`.

src/actions/index.js
```diff
@@ -560,180 +567,178 @@ function updateOutput(
...
-            dispatch(updateProps(observerUpdatePayload));
+            const {paths} = getState();
 
-            dispatch(
-                notifyObservers({
-                    id: outputComponentId,
-                    props: data.response.props,
-                })
-            );
+            Object.keys(data.response).map(k => [k, data.response[k]]).forEach(([k, props]) => {
```
T4rk1n: Iterate over each key of the new response payload, they are now the output prop name.

T4rk1n:
A few issues I uncovered while trying to fix the tests failures:

- [ ] `test_removing_component_while_its_getting_updated` callback output to the parent of the Output Id children, removing the checklist that was there before. 
- [ ] Hot reload has broken.

T4rk1n:
I think the Hot-reload breakage is caused by the changes in #108

src/actions/index.js
```diff
@@ -316,9 +324,12 @@ export function notifyObservers(payload) {
              * of a controller change.
              * for example, perhaps the user has hidden one of the observers
              */
+
+            const {paths} = getState();
+
             if (
                 controllersInFutureQueue.length === 0 &&
```
rmarren1: I don't think `controllersInFutureQueue` will be correct here, since the `controllers` variable set above:
```
            const controllers = InputGraph.hasNode(outputIdAndProp)
                 ? InputGraph.dependantsOf(outputIdAndProp)
                 : [];
```
Would be `[]` whenever we are doing multiple outputs.

src/actions/index.js
```diff
@@ -316,9 +324,12 @@ export function notifyObservers(payload) {
...
-                has(outputComponentId, getState().paths) &&
+                any(e => has(e, paths))(outputIds) &&
                 !controllerIsInExistingQueue
```
rmarren1: Same comment here.

src/actions/index.js
T4rk1n: The controllers variable has the required inputs.

rmarren1: Oh I didn't know those comments were posting :smile:  Makes sense after reading below

src/actions/index.js
```diff
@@ -560,179 +576,188 @@ function updateOutput(
...
+                                getState,
+                                requestUid,
+                                dispatch
+                            );
+                        });
+                    }
                 }
+            };
+            if (multi) {
+                Object.entries(data.response).forEach(handleResponse)
```
rmarren1: could use ramda `forEach`

src/actions/index.js
```diff
@@ -390,18 +397,28 @@ function updateOutput(
      * }
      *
      */
+
+    const [outputComponentId, outputProp] = outputIdAndProp.split('.');
```
rmarren1: I thought this was a mistake at first, but it looks like in the multi-output case it does not matter that these variables are incorrect. Perhaps only create these locals when `config.multi_output` for readability?

src/actions/index.js
```diff
@@ -242,7 +242,15 @@ export function notifyObservers(payload) {
         );
         const queuedObservers = [];
         outputObservers.forEach(function filterObservers(outputIdAndProp) {
```
rmarren1: I found the variable name `outputIdAndProp` confusing when reading this, since it can now refer to stuff like `[output1.children:output2.n_clicks]`

src/actions/index.js
```diff
@@ -390,18 +397,28 @@ function updateOutput(
...
     if (event) {
         payload.event = event;
     }
 
     const {inputs, state} = dependenciesRequest.content.find(
-        dependency =>
-            dependency.output.id === outputComponentId &&
-            dependency.output.property === outputProp
+        dependency => {
+            if (config.multi_output) {
```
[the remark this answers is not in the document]
T4rk1n: Yes, the renderer can be updated without dash.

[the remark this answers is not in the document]
rmarren1: Okay, then if we lock `dash-renderer` version in `dash` install requires then we should be good.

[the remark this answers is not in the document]
T4rk1n: Yes

chriddyp:
This is really awesome stuff. I'll leave it to others to dig into the code and do a full review, but from a high level I'd really like to see an extensive suite of high-level tests for different arrangements of the dependencies (DAG), timing issues, hide/showing component behaviour, and callback dependency "chains". The set of tests in https://github.com/plotly/dash-renderer/blob/master/tests/test_render.py were the only way that I've maintained sanity while working through refactors and bug fixes.

I've outlined some pseudocode of the types of tests that I would be writing to make sure that all of this tricky behaviour is locked down. There are probably many other scenarios to consider, but hopefully this gets everyone's wheels turning :)


1. `test_single_input_multi_outputs_on_multiple_components`

A super simple 1-Many case. Also pushing up the number of outputs past 10 as sometimes there can be weird bugs between 9 and 10 (from one digit to two digits).

```python
N_OUTPUTS = 50

app.layout = html.Div(
    [dcc.Input(id='input', value='dash')] +
    [html.Div(id='output-{}'.format(i)) for i in range(N_OUTPUTS)]
])

@app.callback([Output('output-{}'.format(i), 'children') for i in range(N_OUTPUTS)], [Input('input', 'value')])
def update_output(value):
    call_count.value += 1
    return ['{} - {}'.format(i, value) for i in range(N_OUTPUTS)]

# pseudo-testing code here, but you get the idea
for i in range(N_OUTPUTS):
    self.assertTextEqual(
        'output-{}'.format(i),
        '{} - dash'.format(i)
    )

self.assertEqual(call_count.value, 1)

sendkeys('input', ' hello')
for i in range(N_OUTPUTS):
    self.assertTextEqual(
        'output-{}'.format(i),
        '{} - dash hello'.format(i)
    )

self.assertEqual(call_count.value, 2)
```

2. `test_multi_outputs_on_single_component`

This one is a sanity test that multiple properties on the same component will get updated.

```python
app.layout = html.Div([
    dcc.Input(id='input', value='dash'),
    html.Div(id='output'),
])

@app.callback(
    [Output('output', 'children'),
     Output('output', 'style'),
     Output('output', 'className')],

    [Input('input', 'value')])
def update_output(value):
    call_count.value += 1
    return [
        value,
        {'fontFamily': value},
        value
    ]


# pseudo-testing code here, but you get the idea
self.assertHTMLEqual(
    'output',
    '<div class="dash" style="font-family: dash">dash<div>'
)
self.assertEqual(call_count.value, 1)

sendkeys('input', ' hello')
self.assertHTMLEqual(
    'output',
    '<div class="dash hello" style="font-family: dash hello">dash hello<div>'
)
self.assertEqual(call_count.value, 2)
```

3. `test_returning_nested_components_that_trigger_other_callbacks`

This one tests layouts that generate other components and, when generated, trigger other callbacks. This one also swaps out entire trees, to make sure that all of that state management is clean.

```python
app.layout = html.Div([
    dcc.RadioItems(
        id='input',
        options=[{'label': i, 'value': i} for i in ['tree', 'string']]
        value='tree'
    ),
    html.Div(id='output-1'),
    html.Div(id='middle', children='text in the middle'),
    html.Div(id='output-2'),
])

@app.callback(
    [Output('output-1', 'children'),
     Output('output-2', 'children')],

    [Input('input', 'value')])
def update_output(value):
    call_count.value += 1
    if value == 'tree':
        return [
            html.Div(id='output-1-child', children=[
                html.Div(
                    id='output-1-grandchild-1',
                    children='output 1 grandchild 1'
                ),
                html.Div(
                    id='output-1-grandchild-2',
                    children='output 1 grandchild 2'
                ),
                dcc.Input(id='output-1-input', value='initial value 1'),
                html.Div(id='output-1-grandchild-3')
            ]),

            html.Div(id='output-2-child', children=[
                html.Div(
                    id='output-2-grandchild-1',
                    children='output 2 grandchild 1'
                ),
                html.Div(
                    id='output-2-grandchild-2',
                    children='output 2 grandchild 2'
                ),
                dcc.Input(id='output-2-input', value='initial value 2'),
                html.Div(id='output-2-grandchild-3'),
                html.Div(id='output-2-grandchild-4'),
                html.Div(id='output-2-grandchild-5'),
            ]),

        ]

    else:
        return ['output 1 text', 'output 2 text']


@app.callback(
    [Output('output-1-grandchild-3', 'children'),
     Output('output-2-grandchild-3', 'children')],
    [Input('output-1-input', 'value'), Input('output-2-input', 'value')])
def callback1(value_1, value_2):
    call_counts['callback1'].value += 1
    return [
        'output 1.3 - {} - {}'.format(value_1, value_2),
        'output 2.3 - {} - {}'.format(value_1, value_2),
    ]


@app.callback(
    [Output('output-2-grandchild-4', 'children'),
     Output('output-2-grandchild-5', 'children')],
    [Input('output-2-input', 'value')])
def callback2(value):
    call_counts['callback2'].value += 1
    return [
        html.Div('ggc0 - output 2.4 - {}'.format(value), id='great-grandchild-0')
        html.Div([
            dcc.Input(id='great-grandchild-1', value='ggc1'),
            [html.Div(id='great-grandchild-2'),
             html.Div(id='great-grandchild-3')]
        ])
    ]

@app.callback(
    [Output('great-grandchild-2', 'children'),
     Output('great-grandchild-3', 'children')],
    [Input('great-grandchild-1', 'value')])
def callback3(value):
    call_counts['callback3'].value += 1
    return [
        'ggc2 - {}'.format(value),
        'ggc3 - {}'.format(value),
    ]

def tree_assertions():
    self.assertTextEqual('middle', 'text in the middle')
    self.assertTextEqual('output-1-grandchild-1', 'output 1 grandchild 1')
    self.assertTextEqual('output-1-grandchild-2', 'output 1 grandchild 2')
    self.assertTextEqual('output-2-grandchild-1', 'output 2 grandchild 1')
    self.assertTextEqual('output-2-grandchild-2', 'output 2 grandchild 2')

    self.assertTextEqual('output-1-grandchild-3', 'output 1.3 - initial value 1 - initial value 2')
    self.assertTextEqual('output-2-grandchild-3', 'output 2.3 - initial value 1 - initial value 2')
    self.assertTextEqual('output-2-grandchild-4', 'output 2.4 - initial value 2')
    self.assertTextEqual('great-grandchild-0', 'ggc0 - output 2.4 - initial value 2')
    self.assertTextEqual('great-grandchild-1', 'ggc1')
    self.assertTextEqual('great-grandchild-2', 'ggc2 - ggc1')
    self.assertTextEqual('great-grandchild-3', 'ggc3 - ggc1')

    self.assertEqual(call_counts['callback1'], 1)
    self.assertEqual(call_counts['callback2'], 1)
    self.assertEqual(call_counts['callback3'], 1)

    sendkeys('output-1-input', '!')
    self.assertTextEqual('output-1-grandchild-3', 'output 1.3 - initial value 1! - initial value 2')
    self.assertTextEqual('output-2-grandchild-4', 'output 2.3 - initial value 1! - initial value 2')
    self.assertEqual(call_counts['callback1'], 1)
    self.assertEqual(call_counts['callback2'], 2)
    self.assertEqual(call_counts['callback3'], 1)

    sendkeys('output-2-input', '?')
    self.assertTextEqual('output-1-grandchild-3', 'output 1.3 - initial value 1! - initial value 2?')
    self.assertTextEqual('output-2-grandchild-4', 'output 2.3 - initial value 1! - initial value 2?')
    self.assertTextEqual('output-2-grandchild-4', 'output 2.4 - initial value 2')
    self.assertTextEqual('great-grandchild-0', 'ggc0 - output 2.4 - initial value 2?')
    self.assertTextEqual('great-grandchild-1', 'ggc1')
    self.assertTextEqual('great-grandchild-2', 'ggc2 - ggc1')
    self.assertTextEqual('great-grandchild-3', 'ggc3 - ggc1')    
    self.assertEqual(call_counts['callback1'], 1)
    self.assertEqual(call_counts['callback2'], 3)
    self.assertEqual(call_counts['callback3'], 2)

    sendkeys('great-grandchild-1', '$')
    self.assertTextEqual('great-grandchild-2', 'ggc2 - ggc1$')    self.assertTextEqual('great-grandchild-3', 'ggc3 - ggc1$')
    self.assertEqual(call_counts['callback1'], 1)
    self.assertEqual(call_counts['callback2'], 3)
    self.assertEqual(call_counts['callback3'], 3)

    # reset for the next call
    call_counts['callback1'] = 0
    call_counts['callback2'] = 0
    call_counts['callback3'] = 0

def string_assertions(callcounts):
    self.assertTextEqual('output-1', 'output 1 text')
    self.assertTextEqual('output-2', 'output 2 text')

tree_assertions(1)
click_on_radio_items('input', 'string')
string_assertions()
click_on_radio_items('input', 'tree')
tree_assertions()
click_on_radio_items('input', 'string')
string_assertions()
```

4. Callbacks triggering other callbacks.

This set of tests test that multiple outputs can trigger other multiple outputs. There are lots of different DAG shapes that we should consider here (diamonds vs straight lines, etc).

For these, I recommend just drawing out different DAG shapes on paper. Here are some examples of different shapes (each newline represents a different callback)

**shape 1 - diamond**
```
A -> [B, C]
[B, C] -> D
```

**shape 2 - grandparent hooking in**
```
A -> [B, C]
[A, B, C] -> D
```
(D shouldn't update until B and C are finished updating)

**shape 3 - multi-gen**
```
A -> [C, E]
B -> [D, F]
[C, D, E, F] -> G
```
(G shouldn't update until [C, E] _and_ [D, F] have updated)

**shape 4 multi-gen with granparent hook**
```
A -> [C, E]
B -> [D, F]
[A, B, C, D, E, F] -> G
```

**shape 5 - multi-gen fan out**
```
A -> [B, C]
B -> [D, E]
C -> [F, G]
```

+ there's probably many more unique shapes to consider!

Here's an example of how the test for shape 1 might be written.

```python
call_counts = {
    'bc': Value('i', 0),
    'd': Value('i', 0),
}

app.layout = html.Div([
    dcc.Input(id='a', value='a'),
    dcc.Input(id='b'),
    dcc.Input(id='c'),
    dcc.Input(id='d'),
])

@app.callback([Output('b', 'value'), Output('c', 'value')],
              [Input('a', 'value')])
def update_b_c(value):
    call_counts['bc'].value += 1
    return ['b - ({})'.format(value), 'c - ({})'.format(value)]

@app.callback(Output('d', 'value'),
              [Input('b', 'value'), Input('c', 'value')])
def update_d(valueb, valuec):
    call_counts['d'].value += 1
    return 'd - ({}) - ({})'.format(valueb, valuec)


wait_for_text_to_equal('#d', 'd - (b - (a)) - (c - (a))')
wait_for_text_to_equal('#b', 'b - (a)')
wait_for_text_to_equal('#c', 'c - (a)')
self.assert_equal(call_counts['bc'].value, 1)
self.assert_equal(call_counts['d'].value, 1)

sendkeys('b', '!')
wait_for_text_to_equal('#d', 'd - (b - (a)!) - (c - (a))')
wait_for_text_to_equal('#b', 'b - (a)!')
wait_for_text_to_equal('#c', 'c - (a)')
self.assert_equal(call_counts['bc'].value, 1)
self.assert_equal(call_counts['d'].value, 2)

sendkeys('a', '?')
wait_for_text_to_equal('#d', 'd - (b - (a?)!) - (c - (a?))')
wait_for_text_to_equal('#b', 'b - (a?)!')
wait_for_text_to_equal('#c', 'c - (a?)')
self.assert_equal(call_counts['bc'].value, 2)
self.assert_equal(call_counts['d'].value, 3)
```

5. Timing issues
There is a fair amount of logic that is used to reject old requests. There should be at least one test that makes sure that this logic still works with multiple outputs, see https://github.com/plotly/dash-renderer/blob/657ab923f446212af205d18ba079c8c2ffdfc9ab/tests/test_render.py#L1490-L1525.

6. Hiding components during updates
What happens when a callback returns with an update but a different callback removed that component? This should work with multiple outputs as well, see the single output case here: https://github.com/plotly/dash-renderer/blob/657ab923f446212af205d18ba079c8c2ffdfc9ab/tests/test_render.py#L1297

T4rk1n:
@alexcjohnson The circular dependency test is now passing.

> is a second dep graph that expands out the multiple outputs

That was the winner solution. :tada:

T4rk1n:
There's an error in the style tag of the `.Select` class that is inserted in head and sent to percy; It contains `&gt;`  in some place instead of spaces so the css appear broken.

alexcjohnson:
> There's an error in the style tag of the `.Select` class that is inserted in head and sent to percy; It contains `&gt;` in some place instead of spaces so the css appear broken.

Ooh interesting - how did you figure this out? I guess it's the same as in #120? cc @Marc-Andre-Rivet

alexcjohnson:
> There's an error in the style tag of the `.Select` class that is inserted in head and sent to percy;

And the next question, what would it take to fix that? Is it our error, or is it in `react-select` or one of its neighbors?

T4rk1n:
I downloaded percy sources, open the raw html in a browser, there's no css applied to dropdown unlike when running the test.

In the editor I then see red:
```html
</style><style type="text/css">.Select,.Select-control{position:relative}.Select-control,.Select-input&gt;input
```

T4rk1n:
> And the next question, what would it take to fix that? Is it our error, or is it in `react-select` or one of its neighbors?

It is our error, before the style was in a style sheet, now it's included in the head, there must something that escape the css `.Select > input`.

T4rk1n:
The `>` character is only escaped when it is sent to percy, I contacted percy support.

alexcjohnson:
Thanks @T4rk1n - let’s accept these broken images for now and follow up when we learn more from Percy in https://github.com/plotly/dash-core/issues/43

src/actions/index.js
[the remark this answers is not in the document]
alexcjohnson: Is there a reason to keep `config.multi_output` now that version locking is done?

alexcjohnson:
(I still think we would do well to remove `config.multi_output` for better maintainability, but in the interest of keeping this moving, I'd be OK with just making an issue to remind us to do that later)

PR #93, opened 2018-10-30
Loading states api

valentijnnieman:
Here's the first pass at a loading states api, where `dash-renderer` now provides a `loading_state` prop to components that are taking some time to load. It is used with a dist of [this PR](https://github.com/plotly/dash-core-components/pull/352/) that provides a `Loading` component, that can wrap any other Dash component, making it display a loading spinner if it's not fully rendered. ~~An example is included in `simple.py`, make sure to install the `dev-requirements.txt` before running it!~~

TODO:

- [x] Include which property is loading. Currently we strip the component's name, but it also includes the property that's loading, so we could pass that down as well.
- [x] Clean-up!

The `loading_state` prop is an object that looks like: 

```
{
    is_loading: bool,
    component_name: string,
    prop_name: string
}
```

So that component authors can determine when to show a loading spinner, and know which component in the chain caused the loading to happen (it will return the `id` of the component), and which prop is loading (for example `children`).

Community post: https://community.plot.ly/t/loading-states-api-and-a-loading-component-prerelease


src/APIController.react.js
```diff
@@ -97,14 +95,26 @@ class UnconnectedContainer extends Component {
...
-
-        return <div className="_dash-loading">{'Loading...'}</div>;
+        if (isEmpty(layout)) {
+            return (
+                <div id="_dash-app-content">
+                    <TreeContainer layout={layout} loading={true} />
+                </div>
+            );
+        }
+        window.console.log('return at end of render()');
```
T4rk1n: :zap:

valentijnnieman: Yeah, this isn't ready for review yet, still WIP!

T4rk1n: I was just browsing to see how it would go with #92 and #91 and noticed that.

chriddyp:
For those following along in the community, here's a demo:
image
loading states

src/components/core/NotifyObservers.react.js
```diff
@@ -56,9 +60,30 @@ function NotifyObserversComponent({
...
+    // loading prop coming from TreeContainer
+    let isLoading = loading;
+    let loadingProp;
+    let loadingComponent;
+
+    requestQueue.map(r => {
+        if (r.status === 'loading' && contains(id, r.controllerId)) {
+            isLoading = true;
+            loadingComponent = r.controllerId.split('.')[0];
+            loadingProp = r.controllerId.split('.')[1];
```
T4rk1n: Can split only one time: `[loadingProp, loadingComponent] = r.controllerId.split('.')`. Also wouldn't we only have the last loadingProp and loadingComponent in the requestQueue if there's multiple callback updating ?

T4rk1n: Had a hard time figuring this out, that is not a proper map. It should be a for loop or a filter or a find.

nicolaskruchten: the most appropriate would be a `.forEach()`

src/components/core/NotifyObservers.react.js
```diff
@@ -56,9 +60,30 @@ function NotifyObserversComponent({
...
+        }
+    });
+
+    const thisRequest = requestQueue.filter(r => contains(id, r.controllerId));
```
T4rk1n: The filter is here.

src/components/core/Reloader.react.js
```diff
@@ -51,7 +51,7 @@ class Reloader extends React.Component {
                 ) {
                     // Look if it was a css file.
                     let was_css = false;
-                    for (let a of reloadRequest.content.files) {
+                    for (const a of reloadRequest.content.files) {
```
T4rk1n: I think that is an error, make sure it works because I kinda remember putting const and it failed when there was multiple files. There can be multiple files if multiple assets files are saved at the same time.

rmarren1:
Did you mean to include `dash_core_components-0.38.0rc1.tar.gz` binary?

rmarren1:
Also not sure `dash_renderer/favicon.ico` should be there, `favicon` is in `dash`

src/TreeContainer.js
```diff
@@ -78,9 +85,9 @@ function render(component) {
         ...children
     );
 
-    return <NotifyObservers id={componentProps.id}>{parent}</NotifyObservers>;
+    return <NotifyObservers id={componentProps.id} loading={loading}>{parent}</NotifyObservers>;
 }
 
-render.propTypes = {
+recursivelyRender.propTypes = {
```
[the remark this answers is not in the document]
valentijnnieman: What does 🐱 mean?

src/components/core/Reloader.react.js
valentijnnieman: This must be some artifact from a merge with master, I haven't touched any of the reloading code as far as I remember.

valentijnnieman: I haven't had a chance to rebase yet, because I'm still working on fixing the tests

T4rk1n: I think it's prettier that changes it, I remember it changed it also when I ran it, maybe add an exclude here.

src/TreeContainer.js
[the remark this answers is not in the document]
valentijnnieman: Ah never mind, I found [this](https://github.com/plotly/dev-docs/blob/master/basics/code-review-checklist.md#code-review-animals) again 😸

valentijnnieman:
> Did you mean to include `dash_core_components-0.38.0rc1.tar.gz` binary?

Sorry for the messy commits - at first I did have that tarball in, but because it includes updates to the Input component that cause the tests on here to fail, I've taken it out again. 

I'm still trying to figure out why the tests are failing, so if some of you who've reviewed want to help out with that, would love some help!

T4rk1n:
The failures with the input is the same as in https://github.com/plotly/dash-core-components/pull/384.

Running the failing test locally I see this error:
```
 error      TypeError: Cannot read property 'indexOf' of null
    at _indexOf (_indexOf.js?d014:7)
    at _contains (_contains.js?ab11:5)
    at f2 (_curry2.js?5a7c:25)
    at eval (NotifyObservers.react.js?7924:75)
    at Array.map (<anonymous>)
    at NotifyObserversComponent (NotifyObservers.react.js?7924:74)
    at r.render (react-dom@15.4.2.min.js?v=0.15.0&m=1529003453:13)
    at p._renderValidatedComponentWithoutOwnerOrContext (react-dom@15.4.2.min.js?v=0.15.0&m=1529003453:13)
    at p._renderValidatedComponent (react-dom@15.4.2.min.js?v=0.15.0&m=1529003453:13)
    at p._updateRenderedComponent (react-dom@15.4.2.min.js?v=0.15.0&m=1529003453:13)
```

T4rk1n:
There is an error in that test caused by your changes in NotifyObserver

valentijnnieman:
Tests are fixed and I've made updates according to the feedback - could someone re-review and approve if happy with the changes?

src/components/core/Reloader.react.js
T4rk1n: Still need to change this back.

chriddyp:
Could we publish a pre-release on this and the associated packages? That'll make it easier for me, and some other community members, to try out

valentijnnieman:
@chriddyp Sure. I'll publish a prerelease of this and of the [Loading component](https://github.com/plotly/dash-core-components/pull/352) - anything else I need to do? Dash uses the latest version of dash-renderer I think, but I'm not sure if that applies to prereleases too.

valentijnnieman:
`dash-renderer` version `0.16.0rc1` should be up.

chriddyp:
Perfect, thanks! I don't think that anything else needs to be done w.r.t. prerelease publishing. I'll include this in this week's dash-club newsletter as well.

chriddyp:
@valentijnnieman - Could you also create a community forum thread with installation instructions, a few examples, and a couple of GIFs? Here are a few examples of threads that share prerelease versions https://community.plot.ly/t/exploring-a-transitions-api-for-dcc-graph/15468, https://community.plot.ly/t/preserving-ui-state-like-zoom-in-dcc-graph/15793

valentijnnieman:
@chriddyp How will people be able to use this prerelease? I'd like to make a community post, but I don't know how to use a release candidate of `dash-renderer` in a Dash app, since that seems to be installed from within the `dash` package.

T4rk1n:
@valentijnnieman 

They will need to specify the rc version when doing `pip install`, like:

`pip install dash-renderer==0.16.0rc1`
`pip install dash-core-components==0.41.0rc1`

Marc-Andre-Rivet:
@valentijnnieman Could you provide the link to the community examples? Thanks.

valentijnnieman:
@Marc-Andre-Rivet It's in the top comment (edited :) )

dash_renderer/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.16.1'
+__version__ = '0.17.0rc1'
```
Marc-Andre-Rivet: Shouldn't this be 0.17.0?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "dash-renderer",
-  "version": "0.16.1",
+  "version": "0.17.0-rc1",
```
Marc-Andre-Rivet: Shouldn't this be 0.17.0?

Marc-Andre-Rivet:
Would like to see the following issues opened as follow ups:
- enabling direct support of loading state in DCC components (loading_state + animation type)
- integration testing support for dash-renderer (our tests are all at the behavioural / dash server level atm) -- test loading states, test loading states transition, test more complex graph shapes (Loading component with multiple children, Loading comp within a parent loading comp)

Marc-Andre-Rivet:
As discussed with @valentijnnieman, we are currently going for a mixed approach, supporting both
- loading state as a data-attribute on the components + styleguide / css styling of loading comps
- Loading component with 1 level deep `loading_state` support -- the user becomes responsible for putting items that are actually slow to load (e.g. no listening to `children` prop in Dash)

While the updated implementation looks promising. My main concern right now has to do with the mechanics of getting this to work -- the TreeContainer or the NotifyObserver needs to be hooked up to the store's requestQueue through a prop in order to determine the component's state -- the queue is impacted by user actions and changes often during back-and-forth with Dash. Since not all components are pure in html and dcc, or protected by a `componentShouldUpdate` method, this may cause a significant performance impact. This should be studied and minimized before the feature can be released / sent back for community testing. It's possible that the impact will be negligible but  if not, we may have to expose the queue data differently so as to not trigger useless re-render.

chriddyp:
> the TreeContainer or the NotifyObserver needs to be hooked up to the store's requestQueue through a prop in order to determine the component's state

I'm a little rusty on the redux details here, but I thought that _any_ dispatch that happens will trigger an update. I don't believe that there isn't anything in `NotifyObservers` that says "If `store.x` changes, then update. But if `store.y` changes, don't update.". Does this match your understanding too or am I forgetting about something? If so, can you point me to the code where this happens?

So, if my understanding is right, the _existing_ code is `dispatching` to update the request queue which is already cause re-rendering. Passing new props through in `NotifyObserversComponent` shouldn't change this.

Code reference: `dispatch(setRequestQueue(`
https://github.com/plotly/dash-renderer/blob/ed0679420ae48028b99a67792d1b56705f96aea1/src/actions/index.js#L328

Connected `NotifyObservers`
https://github.com/plotly/dash-renderer/blob/ed0679420ae48028b99a67792d1b56705f96aea1/src/components/core/NotifyObservers.react.js#L7-L105

Marc-Andre-Rivet:
@chriddyp No, your understanding seems to be correct. Adding the prop from the store does not have the impact I thought it would have. My thinking was that since N.Obs. had a new prop from the store, that would trigger extra renders but that assumption was incorrect. There are a few extra renders in the observer when first starting the app but after that it's a one to one match with previous behavior.

Thanks for the input.

valentijnnieman:
There seems to be an issue here that's causing the `test_confirm_as_children` in `dash-core-components` to fail. Haven't been able to figure this out yet.

valentijnnieman:
Released as prerelease version `0.18.0rc4`

src/TreeContainer.js
```diff
@@ -1,46 +1,65 @@
...
+} from 'ramda';
+import {STATUS} from './constants/constants';
 
-export default class TreeContainer extends Component {
+class TreeContainer extends Component {
     shouldComponentUpdate(nextProps) {
-        return nextProps.layout !== this.props.layout;
+        return (
+            nextProps.layout !== this.props.layout
+        );
```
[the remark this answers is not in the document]
valentijnnieman: I'm not sure about this. The main reason that I haven't added any `nextProps.requestQueue` checks is that it causes tests to fail. I'm guessing that re-rendering upon differences in the `requestQueue` prop causes some behaviour to be different - I'm seeing `test_radio_buttons_callbacks_generating_children` and `test_hot_reload` tests failing.

PR #64, opened 2018-07-29
[WIP] Dash dev tools

rmarren1:
**Release Candidate**: `pip install dash-renderer==0.14.0rc7`
In this PR, we change the Dash app layout from something like:
```
<div id="react-entry-point">
  <div id="_dash-undo-redo">...</div>
  <div id="_dash-app-content">
    ...
    <SomeDashComponent />
    ...
  </div>
</div>
```
To the following
```
<div id="react-entry-point">
  <div id="_dash-undo-redo">...</div>
  <GlobalErrorContainer>
    <GlobalErrorOverlay>
      <div id="_dash-app-content">
        ...
        <ComponentErrorBoundary>
          <SomeDashComponent /> OR <ComponentErrorOverlay><SomeDashComponent /></ComponentErrorOverlay> OR <ComponentDisabledOverlay><SomeDashComponent /></ComponentDisabledOverlay>
        </ ComponentErrorBoundary>
        ...
      </div>
    </ GlobalErrorOverlay>
  </ GlobalErrorContainer>
</div>
```

The new components serve the following purposes:
* `GlobalErrorContainer` -- Container for the `GlobalErrorOverlay`, this component connects to the Redux store and passes state and functions down to the `GlobalErrorOverlay`, and also creates the Werkzueg error pop-up.
* `GlobalErrorOverlay` -- This component renders the main Dash app (`_dash-app-content`) along side a conditionally rendered 'error menu'.
* `ComponentErrorBoundary` -- This component wraps each Dash component in the `TreeContainer` and implements the `componentDidCatch` lifecycle method to catch errors in those Dash components. If there is an error in the component, a `ComponentErrorOverlay` component is displayed. If the component must be disabled (this component has a child in the callback graph that has an error) a `ComponentDisabledOverlay` component is displayed.

This PR also makes an update to the Redux store, adding an `error` state that looks like the following:
```
{
  ...,
  error : {
    frontEnd: [ { myUID: 'fjdsa89r32',
                       myID: 'button',
                       error: <javascript error object>,
                       info: <javascript info object> } ],
    backEnd: { errorPage: "...the html werkzueg page..."}
 }
  ...
}
```

PR #100, opened 2018-11-25
Dash Dev Tools

rmarren1:
Built on https://github.com/plotly/dash-renderer/pull/64.


rmarren1:
Looks like this problem still remains...
> After investigating a bit more the input clear bug, it boils down to the way selenium clear the input. It doesn't send the onChange event for the cleared value so somehow the initial value get backs in. This is not something that would happen if the user cleared the input manually with backspaces or selecting the whole text and deleting. Thus, it is safe to patch the failing tests to fix that selenium bug.

T4rk1n:
Yes it's still there, what is fixed is the callback counts, we need to modify the failing test.

byronz:
@alexcjohnson  @Marc-Andre-Rivet would like to get some feedback on the last two fixes & improvement of unit tests:
1. [test initial state](https://github.com/plotly/dash-renderer/commit/a85afe6)
2. [test test_callbacks_generating_children](https://github.com/plotly/dash-renderer/commit/a51bf89)

it demonstrates the direction to make test case more concise and robust, but the current enhancements are still not ideal as the scope of each test is still big and lack of granularity.

nicolaskruchten:
Let's remember to add some tests for the dev bundle :)

nicolaskruchten:
possibly we should add some tests to the prod bundle that ensure that it's not "too big" ?

PR #101, opened 2018-11-27
Add key to rendered components, so React doesn't optimize

valentijnnieman:
This is an interesting fix, where we're now passing a `key` to the rendered components, so that React doesn't try to optimize and recognizes that the components are different. This is an issue for having multiple Graph components for example, where if there's no key, React will try to optimize and will use the same state for both Graphs. This is what's happening in [dcc #379](https://github.com/plotly/dash-core-components/issues/379), and this PR will fix that.


chriddyp:
Can we add an integration test?

valentijnnieman:
@chriddyp I didn't add any integration tests here because these are React internals - what should it test? Should it test that the React component's constructor() is called for each new Graph component? Or test that React passes a key prop correctly to the components? I guess I can add a test that tests if click data is being passed through to a callback, when there's multiple Graph components, but that feels more like a DCC test to me.

chriddyp:
>  I guess I can add a test that tests if click data is being passed through to a callback, when there's multiple Graph components, but that feels more like a DCC test to me.

That's what I'd recommend testing. Basically, we have an example that is broken right now (tabs in a graph) because of dash-renderer. This PR proposes that it fixes this example, so we should add a test that mimics that example to "prove" that it indeed fixes the issue.

chriddyp:
> Or test that React passes a key prop correctly to the components?

This in and of itself isn't important for me to test as we have to _infer_ that adding a key will prevent something from using the same instance which will prevent odd event handling bug code like this. So, I'd rather test a step further back with the actual tabs and graphs.

src/TreeContainer.js
```diff
@@ -78,7 +78,7 @@ function render(component) {
         ...children
     );
 
-    return <NotifyObservers id={componentProps.id}>{parent}</NotifyObservers>;
+    return <NotifyObservers key={componentProps.id} id={componentProps.id}>{parent}</NotifyObservers>;
```
T4rk1n: What if components doesn't have an id ?

valentijnnieman: Well, then the component won't have a `key` either. In the case of the `Graph` component there's always an id - because if there isn't one defined it generates a random one. We could generate something here if id isn't set, or we could make sure that all components that are having trouble with shared state always have an id. Not sure what makes more sense.

T4rk1n: I had tried to make a key generator with the position in the array of children and type or the id if available. The key looked like:

`Tabs/0/Tab/#output`

I don't know if I still have that code, it was in the component as props PR but it didn't solve what I wanted so I may have stripped it out.

chriddyp:
Could we reorder the commits so that the test is first? That way we can make sure that the test is failing on CircleCI and then is subsequently fixed with the next commit. Once that it is done, 💃

src/TreeContainer.js
chriddyp: Could we be more explicit about only providing the key if `id` is defined? Right now, we'll be setting it to `undefined` which might work, but I think it'd be preferrable to just not set it if it's not defined. Otherwise, we'll likely get warnings like "Key is undefined which isn't valid" (https://github.com/facebook/react/issues/11935)

tests/test_render.py
```diff
@@ -45,6 +45,21 @@ def wait_for_text_to_equal(self, selector, assertion_text):
 
         raise exception
 
+    def wait_for_text_to_not_equal(self, selector, assertion_text):
+        start_time = time.time()
+        exception = Exception('Time ran out, {} on {} not found'.format(
+            assertion_text, selector))
+        while time.time() < start_time + 20:
```
chriddyp: There are bulit-in wait-for tools provided by selenium that we should be using instead of doing this timing logic ourselves: https://selenium-python.readthedocs.io/waits.html

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+
+        self.startServer(app)
+
+        self.wait_for_element_by_css_selector('#graph1')
+
+        (self.driver.find_elements_by_css_selector(
+            '#graph1'
+        )[0]).click()
+
+        self.wait_for_text_to_not_equal('#graph1_info', 'null')
```
chriddyp: could we explicitly match the text instead of testing that it's not null? That way, we can ensure that the text being displayed from graph1 is different from the text being displayed in graph2.

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+            else:
+                return tab1_layout
+
+        self.startServer(app)
+
+        self.wait_for_element_by_css_selector('#graph1')
+
+        (self.driver.find_elements_by_css_selector(
+            '#graph1'
+        )[0]).click()
```
chriddyp: I'm surprised that clicking on the graph causes `clickData` to fire, as `clickData` is only bound to the points of the graph, not the graph itself

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+            '#tab2'
+        )[0]).click()
+
+        self.wait_for_element_by_css_selector('#graph2')
+
+        (self.driver.find_elements_by_css_selector(
+            '#graph2'
+        )[0]).click()
+        
+        self.wait_for_text_to_not_equal('#graph2_info', 'null')
```
chriddyp: could you update your editor to always add a newline after saving a file? otherwise, the next person that makes a change to this file will have to commit the newline themselves

src/TreeContainer.js
valentijnnieman: I think React won't add the key prop if it's assignment is undefined, but now that you mention it I'm not sure. Will check this out.

tests/test_render.py
valentijnnieman: Note that the text output on click is very specific, and I was considering that it might change at some point, so that this test would fail every time the click data looks a little different.

valentijnnieman: (which is why I thought maybe asserting that the text isn't null - which is what's happening in that bug - would be a better solution)

tests/test_render.py
valentijnnieman: Good point, although I just copied the function from above - perhaps we can make a separate PR that fixes up these integration tests with the proper Selenium api's etc.

tests/test_render.py
chriddyp: I think this text is very specific and not subject to change. If it does change, then it's like a breaking change (unless its additive).

tests/test_render.py
chriddyp: Sure, feel free to add it to linting too. But note that this is very standard behaviour that we use across all projects.

tests/test_render.py
chriddyp: Sure, but I often find that if we do it in a separate PR, it frequently doesn't get done. I'd much rather do things the Right Way the first time.

src/TreeContainer.js
chriddyp: Feel free to dig in, but this would be a very simple fix that would also improve readability. Otherwise, folks will always wonder "What happens if this is undefined?" when they read the code.
```
extra_props = {}
if (R.has('id', props)) {
   extra_props.key = props.id;
}
return <NotifyObservers {...extra_props}/>
```

tests/test_render.py
chriddyp: In other words, it's just a good default to have in your editor for all projects, whether they have linting in there or not: https://stackoverflow.com/a/5813359/4142536

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+                                        }]
+                                })]),
+
+            html.Pre(id='graph2_info'),
+        ]
+
+        @app.callback(Output(component_id='graph1_info', component_property='children'),
+                    [Input(component_id='graph1', component_property='clickData')])
+        def display_hover_data(hover_data):
+            print(hover_data)
```
chriddyp: ⚡️ remove these print statements too

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+            show_well_info = json.dumps(hover_data)
```
chriddyp: Looks like this was copied and pasted, `show_well_info` isn't meaningful here, can we rename these to like `json_hover_data` or just `return json.dumps(hover_data)`

chriddyp:
Since this is taking a little longer to get through review, could we create a pre-release so that any community members or customers that are affected by this bug can try out the prerelease?

tests/test_render.py
valentijnnieman: Well apparently it's not a default in VSCode. I'm fine with toggling that switch ofcourse, just saying it might be easier to enforce those rules in linting if you feel strongly about them and if it is considered standard behaviour for all of our projects.

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+            elif tab == "tab2":
+                return tab2_layout
+            else:
+                return tab1_layout
+
+        self.startServer(app)
+
+        self.wait_for_element_by_css_selector('#graph1')
+
+        (self.driver.find_elements_by_css_selector(
```
T4rk1n: :zap: unnecessary parens

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+
+        self.wait_for_text_to_not_equal('#graph1_info', 'null')
+
+        (self.driver.find_elements_by_css_selector(
+            '#tab2'
+        )[0]).click()
+
+        self.wait_for_element_by_css_selector('#graph2')
+
+        (self.driver.find_elements_by_css_selector(
```
T4rk1n: :zap: unnecessary parens

tests/test_render.py
```diff
@@ -1866,3 +1881,94 @@ def update_output(*args):
...
+
+        self.wait_for_element_by_css_selector('#graph1')
+
+        (self.driver.find_elements_by_css_selector(
+            '#graph1'
+        )[0]).click()
+
+        self.wait_for_text_to_not_equal('#graph1_info', 'null')
+
+        (self.driver.find_elements_by_css_selector(
```
T4rk1n: :zap:

tests/test_render.py
valentijnnieman: Since I just copy-pasted the function above, I feel asking me to refactor these functions is a bit out of scope - which is why I suggest doing it in another PR, or creating an issue for this. I'd rather see this PR get through a bit quicker than getting bogged down with - although good and reasonable - suggestions to fix other things. Technically, the Right Way could've been done when the function I copy-pasted was first added.

tests/test_render.py
valentijnnieman: Well, what if some data does get added at some point? Then we'd have to fix/update this test too. Wouldn't it be better if we had tests that only fail when there's an actual regression/error?

tests/test_render.py
valentijnnieman: #103 here's an issue for it.

tests/test_render.py
chriddyp: Yup, go ahead and add it to eslint. Just note that not all projects have linting, and so I recommend making it your default in your code editor as well as per our organizational standard practices.

tests/test_render.py
chriddyp: Then we'd fix the test. It's only like 10 characters and you'll just copy and paste it based off the test failure, it will only take 5 minutes. Similar to how we, e.g. update jest tests when jest upgrades.

In the current verison of the test, I can't ensure that the text that your comparing is actually the proper clickData hover data. And from other comments, I'm surprised that the click data would be correct (since we're not actually clicking on points) and also the current test doesn't actually test that the clickData is different between the two graphs - we don't know from the test that the even isn't actually the same _same instance_ of the graph across tabs (I mentioned this in my original comment: https://github.com/plotly/dash-renderer/pull/101#discussion_r237667873).

valentijnnieman: You're right in that we're not testing that at the moment. However, we test that the clickData is no longer null as it was before, proving that the clickData is coming in and that the click event is being bound. Previously, the events weren't being bound since it used the same instance of the first Graph component. But I see your point that it is not 100% convincing evidence. Asserting that Graph1 clickData does not equal Graph2 clickData seems like a good addition, I'll add that!

tests/test_render.py
valentijnnieman: Sure, will make it my default. Do you have a link for me for the organizational standard practices? I'm looking through our dev docs but can't find anything about that.

tests/test_render.py
chriddyp: Please just assert the actual content of `clickData`, not just that they aren't equal to each other. I want to see that the clickData that is coming in is correct.

tests/test_render.py
T4rk1n: For python you should have atleast flake8.

tests/test_render.py
valentijnnieman: Tests are failing on 2.7 because the text is slightly different - there's a `u` in front of there etc. Hmm!

tests/test_render.py
valentijnnieman: If you run the test (or the app used in the test) you see that when you hover over the graph it "focusses" on the first graph point, so looks like clicking there will fire on the closest point.

tests/test_render.py
bpostlethwaite: @valentijnnieman here are some standard practice docs https://github.com/plotly/dev-docs/blob/master/reference/styleguides/python.md.
pep8 does recommend a newline at the end of a file but as @chriddyp suggested it is fairly standard programming practice (across languages and linters) to include a newline terminator https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline. 

So while it _should_ be in a linter it is better to turn it on in your editor options so that you append a newline even in projects that don't have linting.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.16.1] - 2018-11-27
+### Added
```
Marc-Andre-Rivet: Internally this is added -- externally this is a bug that's been fixed. Change to ```### Fixed`

Marc-Andre-Rivet:
> Could we reorder the commits so that the test is first? That way we can make sure that the test is failing on CircleCI and then is subsequently fixed with the next commit.

I agree we should see this test failing without the fix before concluding the fix is correct -- the test is very remote from the code fix and we have to infer a lot. Barring reordering the commits to get the desired failure --> success chronology, just revert the fix (it's just one line after all) for one build and add it back.

In the future we should strive to make the test failure / success more explicit in our PRs.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.16.1] - 2018-11-27
```
Marc-Andre-Rivet:
```suggestion
## [0.16.1] - 2018-12-14
```

PR #108, opened 2018-12-13
Remove `Authentication` component.

T4rk1n:
- Removed the `Authentication` component, all it did was reading the configs.
- Moved the readConfig dispatch to `AppContainer`.


T4rk1n:
I looked back at the Authentication component and found a `_dash-login` request. There's a `_dash-login` route in dash-auth but it's handled by it's oauth.js bundle.

@chriddyp Can you confirm that removing this api will not break anything ?

PR #109, opened 2018-12-17
Dev tools front end

valentijnnieman:
This is a first pass at the UI for the dev tools. This includes:
- Front end error cards
- Front end error cards list
- Debug menu

This does _not_ include:
- Werkzeug debugger styles
- Hot reloading toggle
- Callback graph

What else is still left to do?
- Front end error cards better stack trace - right now there's just a basic stack trace - the designs show a much more detailed stack trace that I am not sure can be implement at this time.
- Lots of style fixes! This is a first pass, and it is slightly rough. 

@rmarren1 Would be great if you could take a look. Wanted to provide this PR so not to possibly block you. I'm on vacation starting tomorrow, will pick this up further in the new year! Happy holidays 🎄 🎅 

I've used the example app you provided me, and the version of html_components you made that included the Button front end errors. Here is the example again:

```import dash

external_stylesheets = [
    'https://codepen.io/chriddyp/pen/bWLwgP.css',
    'https://unpkg.com/tachyons@4.10.0/css/tachyons.min.css'
]

app = dash.Dash(external_stylesheets=external_stylesheets)

app.layout = html.Div(
    html.Div([
        html.Div(
            html.H1("My Sample App"),
            className="row"
        ),
        dcc.Textarea(
            placeholder="This is a sample dash app",
            id='text'
        ),
        html.Div([
            html.Button(id='button1', children='button1', className='mh2 mv4'),
            html.Button(id='button2', children='button2', className='mh2 mv4')
        ], className="row"),
        html.Div([
            html.Button(id='button3', children='simulate back-end error'),
        ], className="row"),
        html.Div(id='dummy-output')
    ], className="jumbotron"),
    className='container text-center'
)


class MyCustomException(Exception):
    pass

@app.callback(Output('dummy-output', 'children'),
              [Input('button1', 'n_clicks'),
               Input('button2', 'n_clicks'),
               Input('button3', 'n_clicks')])
def callback(b1, b2, b3):
    if b3:
        raise MyCustomException("BackEnd Error.")
    return ""

@app.callback(Output('button1', 'children'),
              [Input('text', 'value')])
def callback2(text):
    return 'button1'

@app.callback(Output('button2', 'children'),
              [Input('text', 'value')])
def callback3(text):
    return 'button2'

app.run_server(debug=True, port=8000)
```


valentijnnieman:
Here's a screenshot!:

screen shot 2018-12-17 at 4 21 08 pm

rmarren1:
Wow screenshot looks great! I'll take a test drive before the holidays, and see if there is an alternative to the werkzueg issue

dash_renderer/__init__.py
```diff
@@ -12,8 +12,8 @@
 from .version import __version__
 __file__
 
-_DEFAULT_REACT_VERSION = '16.2.0'
-_REACT_VERSION_TYPES = {'15.4.2', '16.2.0'}
+_DEFAULT_REACT_VERSION = '16.6.3'
```
rmarren1: 16.7.0 was actually released two days ago, if the goal was the most recent version

package.json
```diff
@@ -40,6 +41,7 @@
     "redux": "^3.4.0",
     "redux-actions": "^0.9.1",
     "redux-thunk": "^2.0.1",
+    "svg-inline-react": "^3.1.0",
```
rmarren1: I think there are better ways to load SVG. It looks like this project and `svg-inline-loader` were created as a work-around in the past when React didn't properly support SVG, and it is just being maintained now. (https://github.com/sairion/svg-inline-react/issues/15#issuecomment-264492050).

A quick search looks like importing SVG into React has a messy history, but I think it would be safe to do it like they do in `create-react-app`, e.g. with [svgr](https://github.com/smooth-code/svgr).

src/components/error/FrontEnd/FrontEndErrorContainer.js
```diff
@@ -0,0 +1,56 @@
+import React, {Component} from 'react';
```
rmarren1: Should be named `FrontEndErrorContainer.react.js` for consistency

src/components/error/ComponentErrorOverlay.react.js
```diff
@@ -1,66 +1,29 @@
...
-            <p>{errorString}</p>
-            <button onClick={resolve}>Resolve Error</button>
-          </div>
-        </div>
-      </div>
-    )
-  }
+        return (
+            <div
+                className={'dash-error-overlay'}
```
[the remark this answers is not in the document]
valentijnnieman: That's a good point - I removed the underscore from those prefixes, because in other places such as dash-core-components we don't use them either. Basically it would be more consistent to not use underscores, as that's what other Dash repo's use. `_dash-loading` will also disappear once the Loading component and Loading States are merged. If it's a big issue for you, I can switch everything back to use `_dash`, otherwise I suggest we keep it as-is.

package.json
valentijnnieman: Ah great, I was wondering about that! SVGR looks good - I'll try that out, thanks!

src/components/error/FrontEnd/FrontEndErrorContainer.js
valentijnnieman: 🙇

src/components/error/FrontEnd/FrontEndError.css
```diff
@@ -0,0 +1,59 @@
+.dash-fe-errors {
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: @rmarren1 Could you explain the convention currently used? Does `_dash-*` differ in signification from `dash-*`? My first thought was that the `_` prefix was for donating a private class but `and make it known to users that we are polluting the _dash prefixed namespace` makes me think otherwise.

src/components/error/FrontEnd/FrontEndError.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.dash-fe-error-top {
+    display: flex;
+    align-items: center;
+}
+.dash-fe-error__icon-error {
```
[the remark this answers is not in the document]
Marc-Andre-Rivet: Or maybe it's an icon-error? :)

src/components/error/GlobalErrorContainer.react.js
```diff
@@ -1,68 +1,66 @@
...
+    //     var newWin = open('error.html', 'werkzeug', 'height=1024,width=1280');
+    //     newWin.document.write(error.backEnd.errorPage);
+    //     var debugger_css = newWin.document.getElementsByTagName('link')[0];
+    //     debugger_css.parentNode.removeChild(debugger_css);
+    //     var style = newWin.document.createElement('style');
+    //     style.type = 'text/css';
+    //     style.innerHTML = serverErrorCSS;
+    //     newWin.document.getElementsByTagName('head')[0].appendChild(style);
+    //     newWin.document.close();
+    // }
```
Marc-Andre-Rivet: Is there a reason to keep this around?

src/components/error/GlobalErrorContainer.react.js
```diff
@@ -1,68 +1,66 @@
...
-          </GlobalErrorOverlay>
-        </div>
-      )
-  }
+    render() {
+        const {error, dispatch} = this.props;
+        if(!isEmpty(error.backEnd)) {
+            return <div className='dash-backend-error' dangerouslySetInnerHTML={{__html: error.backEnd.errorPage}}></div>
+        }
+            return (
```
Marc-Andre-Rivet: Indentation

src/components/error/FrontEnd/FrontEndError.css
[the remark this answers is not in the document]
valentijnnieman: That would be the case if there was a `dash-fe-error__icon` base class, but since there isn't, it doesn't make sense to use a modifier (`--`) here I think.

src/components/error/GlobalErrorContainer.react.js
valentijnnieman: Looking back at old code and reminiscing about times past, reminding ourselves how young we really are and how anyone of us can go at any second?

Marc-Andre-Rivet:
@rmarren1 @valentijnnieman We're having an issue here where running the tests against the dev-requirements in `dev-tools` fails for unknown reasons but running them against the version in `master` [works fine](https://github.com/plotly/dash-renderer/pull/109/commits/16a847afbdf2d9f880dbea0dc2a18f5306b9d935).

Based on the nature of the changes, this seems unrelated and seems to link back to the old dash change that caused tests to fail. I will force merge this now.

PR #110, opened 2018-12-19
Fix reload

T4rk1n:
Hot Reload was broken by #108 with the removal of the Authentication component that wrapped the Reloader.

- Keep the config store state on soft reload.
- AppProvider returns `Loading...` if no configs as before #108


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [UNRELEASED] - 2018-12-19
+### Fixed
+[#110](https://github.com/plotly/dash-renderer/pull/110)
+
+- Keep the config store state on soft reload.
+- AppProvider returns `Loading...` if no configs as before #108
+
```
alexcjohnson: Merge all unreleased changes into the same block, right? Since once we *do* make a new release, they will all be part of it.

T4rk1n: I wanted to have the timestamp with the changes. It does look weird, will change.

alexcjohnson:
Looks great 💃 
At some point it might be nice to write a test for hot reload, though offhand that sounds a bit painful...

T4rk1n:
Yes, hot-reload really needs a test, atleast to make sure it still work.

Thinking of something like changing an assets css, wait for reload and assert the style has changed.

PR #114, opened 2019-01-10
Remove events

### LINKED ISSUE #113: Remove Event system
Companion issue to https://github.com/plotly/dash/issues/531

Nothing special is required in `dash-renderer`, just removing the `eventGraph` and related code. But it may be useful to have this prior to multi-output (cc @T4rk1n ) and dynamic callbacks as there may be some additional dependency graph machinery to get those to work right, and it would be nice not to have to duplicate those changes for events.

alexcjohnson:
Closes #113 
Pretty straightforward, just removing the `DepGraph` for events (but note that `state.graphs` is still an object we can add more to, which may be important for multi-outputs and dynamic callbacks to create a modified graph for cycle detection), and all references to `fireEvent`.

As in https://github.com/plotly/dash-html-components/pull/89 I added a test shortcut to package.json - is there already a way to do this that I didn't see?


package.json
```diff
@@ -13,7 +13,8 @@
     "start": "webpack-serve ./webpack.serve.config.js",
     "format": "prettier --config .prettierrc --write src/**/*.js",
     "format:test": "prettier --config .prettierrc src/**/*.js --list-different",
-    "test": "npm run lint"
+    "test": "npm run lint",
+    "test:py": "python -m unittest tests.test_render.Tests && python -m unittest tests.test_race_conditions.Tests"
```
Marc-Andre-Rivet: Please update CircleCI config to use this command instead

Marc-Andre-Rivet:
I think this `Event` should also be removed
https://github.com/plotly/dash-renderer/blob/c21512c88bac80e96b9bb05a481b102303296246/tests/test_race_conditions.py#L6

Marc-Andre-Rivet:
Missing a changelog entry.

Marc-Andre-Rivet:
Getting errors when running dash-docs with this version of dash-renderer.
Seems like one usage of updateOutput in src/actions/index.js was not updated

Need to remove the `null` event. https://github.com/plotly/dash-renderer/blob/c21512c88bac80e96b9bb05a481b102303296246/src/actions/index.js#L704

alexcjohnson:
> Seems like one usage of updateOutput in src/actions/index.js was not updated

Good catch! updated that in https://github.com/plotly/dash-renderer/pull/114/commits/c4a9a22f17695b615e1f377f064156a190bb4d01. It seems like the dash-renderer tests don't touch this unfortunately, but it's a bit tough as I get a few failures with these tests locally even on master. Not going to try and sort that out in this PR, but I'll see if I can add a test that covers that `updateOutput` call.

tests/test_render.py
```diff
@@ -1352,6 +1352,10 @@ def update_body(chapter):
             Output('button-output', 'children'),
             [Input('button', 'n_clicks')])
         def this_callback_takes_forever(n_clicks):
+            if not n_clicks:
+                # initial value is quick, only new value is slow
+                # also don't let the initial value increment call_counts
+                return 'Initial Value'
```
alexcjohnson: I don't understand how this test was passing on master - it failed for me locally, and failed in this branch on ci, until this fix.

tests/utils.py
```diff
@@ -48,19 +52,24 @@ def wrapped_condition_function():
...
+                return True
+        elif condition_val == expected_value:
             return True
         time.sleep(0.5)
 
-    raise WaitForTimeout(get_message())
+    if get_message:
+        message = get_message()
+    elif expected_value:
+        message = 'Final value: {}'.format(condition_val)
```
alexcjohnson: Would have helped me debug `test_removing_component_while_its_getting_updated` - I thought the callback wasn't getting called at all, when in fact it was called too many times! With `expected_value` you can get error messages like:
```
ERROR: test_removing_component_while_its_getting_updated (tests.test_render.Tests)
  File "tests/test_render.py", line 1423, in test_removing_component_while_its_getting_updated
    wait_for(lambda: call_counts['button-output'].value, expected_value=1)
  File "tests/utils.py", line 72, in wait_for
    raise WaitForTimeout(message)
WaitForTimeout: Final value: 2
```
Which would have made it much clearer what was going on.

Would be nice if we could share utils like this across dash repos - perhaps if we split out `dash_development` things like this could go there?

alexcjohnson:
> > Seems like one usage of updateOutput in src/actions/index.js was not updated
> 
> Good catch! updated that in [c4a9a22](https://github.com/plotly/dash-renderer/commit/c4a9a22f17695b615e1f377f064156a190bb4d01). It seems like the dash-renderer tests don't touch this unfortunately, but it's a bit tough as I get a few failures with these tests locally even on master. Not going to try and sort that out in this PR, but I'll see if I can add a test that covers that `updateOutput` call.

Aaaactually we did have a test that covered this case: `test_callbacks_triggered_on_generated_output` - not sure how I missed that the first time, but anyway I made that test a bit clearer in https://github.com/plotly/dash-renderer/pull/114/commits/57ef3b2d4255e70e044aa2821b270842333aa847

Marc-Andre-Rivet:
@alexcjohnson So, I haven't digged enough in renderer to have a fully formed opinion but a few things already come to mind -- obviously nothing in here is in the scope for this PR!

- would very probably benefit from using a typing library
- fine grained testing is HARD when you're only doing integration/e2e tests.. secondary observations used to deduce underlying behavior -- most of the tests of that sort should pbly live in the comps source repo (and potentially be called as additional tests by this repo)
- some tests depend on implementation details of components in dcc and html -- internally defined comps for specfic tests would be better
- def would like to push some shared configurations somewhere else! ts/es/lint configs, version/deployment scripts, etc.

PR #115, opened 2019-01-21
Noevents v1

alexcjohnson:
Merging events and a few other changes from master into release-v1


alexcjohnson:
Still don't know what's up with the percy snapshots having the wrong chapter radio button selected (or the jumble of modebar icons down in the corner, but that's nothing new) but it seems like locally things are correct.

PR #118, opened 2019-02-06
Remove redux logger

### LINKED ISSUE #117: Remove redux-logger?
The redux logger causes a ton of output in the console, especially with hot loading.

Also, sometimes it catches component error messages and rethrows them, making the tracebacks & "launch debugger on uncaught exceptions" tools hard to use.

I'm in favor of removing redux-logger. If anyone benefits from this while developing `dash-renderer`, they can add it back while they are in development (but not commit it).

Here's an associated community report that is having trouble filtering out the logs: https://community.plot.ly/t/dash-renderer-dev-js-logs-messes-my-browser-console/19120

T4rk1n:
Closes #117


package.json
```diff
@@ -54,7 +54,6 @@
     "eslint-plugin-react": "^6.9.0",
     "prettier": "^1.14.3",
     "react-hot-loader": "^1.3.0",
-    "redux-logger": "^2.8.0",
```
chriddyp: does `package-lock.json` need to get updated too?

PR #119, opened 2019-02-06
Fix e2e tests & bump dash version to master

Marc-Andre-Rivet:
Fixes https://github.com/plotly/dash-core/issues/19.
Fixes https://github.com/plotly/dash-core/issues/25.

3 different tests from 3 different dependencies were failing: dash, dcc.input and plotly.js graph

As far as I'm concerned, this is a strong compelling argument for starting writing integration tests for the renderer rather than depending solely/very heavily on e2e tests.


Marc-Andre-Rivet:
Not sure about the 2 remaining Percy diffs -- probably related to the version bump for `react-select` and `react-virtualized-select` in 0.38.1. Will leave it to someone else to confirm this is fine or not for now.

0.43.0 (3.1.3) https://github.com/plotly/dash-core-components/blame/master/package.json#L48
0.38.1 (3.1.0) https://github.com/plotly/dash-core-components/blame/ce6126950178c6ce8a0b14bd682caa3589773568/package.json#L57

@valentijnnieman, insight on this?

.circleci/config.yml
```diff
@@ -30,6 +30,7 @@ jobs:
               python -m venv venv || virtualenv venv
               . venv/bin/activate
               pip install -r dev-requirements.txt --quiet
+              pip install -e .
```
alexcjohnson: Again, not a big deal but I'd think all `install -e` lines should go before all `install -r` (or any other versioned install) lines.

tests/test_render.py
```diff
@@ -486,7 +486,8 @@ def update_output(value):
         self.percy_snapshot(name='simple-callback-1')
 
         input1 = self.wait_for_element_by_css_selector('#input')
-        input1.clear()
+        input1.send_keys(Keys.CONTROL + "a")
```
alexcjohnson: Does that work on a mac, or do we need `Keys.META`?

alexcjohnson:
> Not sure about the 2 remaining Percy diffs

Testing locally, the select box looks fine at that point. I suspect that dynamically-created CSS stylesheets are not being properly replicated by percy - hence the graph modebar icons sitting in a pile at the bottom of the page... Could it be that the new version of `react-select` is doing the same?

tests/test_render.py
```diff
@@ -486,10 +487,13 @@ def update_output(value):
...
 
         input1 = self.wait_for_element_by_css_selector('#input')
-        input1.send_keys(Keys.CONTROL + "a")
-        input1.send_keys(Keys.DELETE);
+        initialValue = input1.get_attribute('value')
 
-        input1.send_keys('hello world')
+        action = ActionChains(self.driver)
+        action.click(input1)
+        action = action.send_keys(Keys.BACKSPACE * len(initialValue))
```
alexcjohnson: Ha ok, super annoying but you're right, neither `META` nor `COMMAND` actually seems to work on a mac. At some point we should probably make a util for clearing an input 🌴 but this is good for now.

PR #120, opened 2019-02-07
Fix e2e tests & bump dash version to master

Marc-Andre-Rivet:
Fixes https://github.com/plotly/dash-core/issues/19


dev-requirements.txt
```diff
@@ -1,4 +1,5 @@
 git+git://github.com/plotly/dash@master#egg=dash
+git+git://github.com/plotly/dash-core-components@fix-tests-in-renderer#egg=dash_core_components
```
alexcjohnson: hmm, who's going to win, this version or the numbered version coming from dash?

Eventually we probably want to move to have CI explicitly clone this (and the other core repos) then build it and `install -e` it, then there's no ambiguity (also we're not dependent on committing the build to the repo 🎉 )

Marc-Andre-Rivet: Did not think this through. And the `pip list | grep dash` line has been removed!

Marc-Andre-Rivet: And that last change won't tell me anything as I left the version number as is

Marc-Andre-Rivet:
So essentially the rendering looks fine in Selenium but broken in Percy.. I have no idea why :( :-1:

Marc-Andre-Rivet:
@alexcjohnson At this point my thinking is as follow: Revert back to fixing the tests, accepting the delta in Percy (manual testing shows that it displays correctly) and winging the Percy investigation in another issue -- we know Percy does this, this is not a new issue, it just happens to be exposed by the dcc version bump.

Marc-Andre-Rivet:
Follow up issue for Percy: https://github.com/plotly/dash-core/issues/43

PR #124, opened 2019-02-14
Send which input fired in update request.

T4rk1n:
https://github.com/plotly/dash/issues/291 & https://github.com/plotly/dash/pull/608


src/actions/index.js
```diff
@@ -356,7 +357,8 @@ function updateOutput(
     outputProp,
     getState,
     requestUid,
-    dispatch
+    dispatch,
+    changedProps,
```
alexcjohnson: Per our slack discussion this morning, I guess change this to `changedPropIds`?

T4rk1n: Yes, also filtering the inputs that are not connected.

T4rk1n:
> Very nice, it's awesome how compact a change this turned out to be!

Turned out almost everything was in place for this in the renderer...

PR #130, opened 2019-03-04
Improve build

Marc-Andre-Rivet:
Run dash-renderer tests against the latest version of dash-*, **always**. When ready this will be copied/applied to the other dash repos.


.circleci/config.yml
```diff
@@ -20,36 +20,42 @@ jobs:
...
+          name: Install dependencies (transient)
+          command: |
+              . venv/bin/activate
+              git clone git@github.com:plotly/dash.git
+              pip install -e ./dash --quiet
+              git clone git@github.com:plotly/dash-core-components.git
+              cd dash-core-components && npm i && cd ..
+              cd dash-core-components && npm run build && cd ..
+              pip uninstall dash-core-components -y
+              pip install -e ./dash-core-components --quiet
```
Marc-Andre-Rivet: Install dash and override dcc

.circleci/config.yml
```diff
@@ -20,36 +20,42 @@ jobs:
...
+
+      - run:
+          name: Build
+          command: |
+              . venv/bin/activate
+              npm run build
+              pip uninstall dash-renderer -y
+              pip install -e . --quiet
+              pip list | grep dash
```
Marc-Andre-Rivet: Build & install dash-renderer, print the installed dash packages, for sanity in case of problems..

requirements-base.txt
```diff
@@ -0,0 +1,4 @@
+percy
+selenium
+mock
+six
```
Marc-Andre-Rivet: CI will only install this

requirements-dash.txt
```diff
@@ -0,0 +1,2 @@
+-e git+git://github.com/plotly/dash.git@master#egg=dash
+-e git+git://github.com/plotly/dash-core-components@fix-select-css#egg=dash-core-components
```
Marc-Andre-Rivet: A default for the local environment -- good enough for most cases

.circleci/config.yml
```diff
@@ -20,36 +20,52 @@ jobs:
...
+              git clone git@github.com:plotly/dash-html-components.git
+              cd dash-html-components && npm i && cd ..
+              cd dash-html-components && npm run build && cd ..
+              pip uninstall dash-html-components -y
+              pip install -e ./dash-html-components --quiet
+              git clone git@github.com:plotly/dash-table.git
+              cd dash-table && npm i && cd ..
+              cd dash-table && npm run build && cd ..
+              pip uninstall dash-table -y
+              pip install -e ./dash-table --quiet
```
Marc-Andre-Rivet: renderer tests were already using html but not asking for them explicitly -- fetch everything and ask questions later..

dev-requirements.txt
```diff
@@ -1,6 +1,2 @@
--e git+git://github.com/plotly/dash.git@master#egg=dash
--e git+git://github.com/plotly/dash-core-components@fix-select-css#egg=dash-core-components
```
Marc-Andre-Rivet: fix-select-css got through the review it seems

.circleci/config.yml
```diff
@@ -20,36 +20,52 @@ jobs:
...
+          command: |
+              . venv/bin/activate
+              git clone git@github.com:plotly/dash.git
+              pip install -e ./dash --quiet
+              git clone git@github.com:plotly/dash-core-components.git
+              cd dash-core-components && npm install --ignore-scripts && cd ..
+              cd dash-core-components && npm run build && cd ..
+              pip uninstall dash-core-components -y
+              pip install -e ./dash-core-components --quiet
+              git clone -b improve-build git@github.com:plotly/dash-html-components.git
```
Marc-Andre-Rivet: `improve-build` NEEDS TO BE REMOVED WHEN https://github.com/plotly/dash-html-components/pull/96 IS MERGED

.circleci/config.yml
Marc-Andre-Rivet: all of this is very verbose right now -- want to wait until the same logic is applied in all of dash-core before refactoring the 🍝

Marc-Andre-Rivet: thinking of a script that takes the repo's name and the branch name in the dash-core repo atm

Marc-Andre-Rivet: the explicit build also brings us a step closer to removing the builds from the repo if desired

dev-requirements.txt
alexcjohnson: yeah I noticed that Friday https://github.com/plotly/dash-core-components/pull/471

requirements-dash.txt
```diff
@@ -0,0 +1,4 @@
+-e git+git://github.com/plotly/dash.git@master#egg=dash
+-e git+git://github.com/plotly/dash-core-components@master#egg=dash-core-components
+-e git+git://github.com/plotly/dash-html-components@master#egg=dash-html-components
+-e git+git://github.com/plotly/dash-table@master#egg=dash-table
```
alexcjohnson: So when would this actually get used? I guess if someone manually runs `pip install -r dev-requirements.txt` because they're editing renderer but not `dash`? If that's all we might not want to encourage this, instead directing people to a more complete dev setup.

.circleci/config.yml
alexcjohnson: Yeah, I think this scheme makes sense. It's certainly easier conceptually - and more flexible, as we have the full dev version of the repo - than trying to do this all via `pip` machinery. It might be nice at some point to do it in a cacheable way - something like check if each repo exists locally; if it does just do a `git pull/checkout` on the appropriate branch; if not, do a `git clone`.

Can you also group all the branch references together, so it's easy to see which ones are on master and which aren't? I guess if there's a script that we just call multiple times, that would take care of this... if not, we could just make vars for the branch names.

Is the `pip uninstall` step really needed if you're just going to do `install -e` right after?

Marc-Andre-Rivet: Yes, I can group things together. I was thinking of scripting this with repo+branch as params later on. I'm not sure about uninstall vs. -e, might just be a bit paranoid again, will test it out.

I can also make master explicit by always using `-b master` even if it's redundant.

requirements-dash.txt
Marc-Andre-Rivet: Maybe I remove this altogether?

.circleci/config.yml
alexcjohnson: After the grouping & simplifying this looks great, probably not even worth adding a script as it'll be easier to follow inline 🏆

.circleci/config.yml
```diff
@@ -20,36 +20,43 @@ jobs:
...
+              cd dash-core-components && npm install --ignore-scripts && npm run build && pip install -e . && cd ..
+              cd dash-html-components && npm install --ignore-scripts && npm run build && pip install -e . && cd ..
+              cd dash-table && npm install --ignore-scripts && npm run build && pip install -e . && cd ..
+
+      - run:
+          name: Build
+          command: |
+              . venv/bin/activate
+              npm run build
+              pip uninstall dash-renderer -y
```
alexcjohnson: oh hmm, you removed `pip uninstall` for the other pieces, can this one go away too?

Marc-Andre-Rivet: Removing

PR #139, opened 2019-03-22
[WIP] Add more hot-reload tests

T4rk1n:
Hot reload broken on master, reload only once when python changes are detected.

- [x] Add test reload twice on assets changes.
- [ ] Add Test reload (twice) on python changes.


byronz:
this can be safely re-addressed in further dash test integration

byronz:
this can be safely re-addressed in further dash test integration

PR #140, opened 2019-03-23
[WIP] Renderer regression

Marc-Andre-Rivet:
To be reviewed alongside https://github.com/plotly/dash-core-components/pull/496.

- [ ] Changelog for previous optimization

Related CI runs for all dash-core projects running against the `regression` branches of dcc and renderer:
https://circleci.com/gh/plotly/dash/3447
https://circleci.com/gh/plotly/dash/3446
https://circleci.com/gh/plotly/dash/3435
https://circleci.com/gh/plotly/dash-core-components/3963
https://circleci.com/gh/plotly/dash-core-components/3962
https://circleci.com/gh/plotly/dash-core-components/3961
https://circleci.com/gh/plotly/dash-html-components/889
https://circleci.com/gh/plotly/dash-html-components/888
https://circleci.com/gh/plotly/dash-html-components/887
https://circleci.com/gh/plotly/dash-table/6283
https://circleci.com/gh/plotly/dash-table/6282
https://circleci.com/gh/plotly/dash-table/6281
https://circleci.com/gh/plotly/dash-table/6280
https://circleci.com/gh/plotly/dash-table/6279
https://circleci.com/gh/plotly/dash-table/6278

This fixes the regression in Loading Component caused by the renderer optimization. The LC was using the nested children tree structure to determine if it was loading or not. Since the tree containers do not contain the child anymore but instead generate them at render, the LC is only capable of reacting to direct children having loading_state=true.

To resolve the loading state of the LC, two information are now required (1) the layout -- accessible through the children and (2) the request queue. A naive implementation adding the request queue to all tree containers and re-rendering on both request queue and layout changes triggers an excessive amount of re-renders, losing a significant portion of the performance gains made through the previous rework.

To remedy this, the renderer now checks for a flag defined on the component `_dashprivate_deepstate` that tells it to determine the state of the component by checking the state of all of its children, pruning when reaching another component with _dashprivate_deepstate set to true. This allows the renderer to determine the loading_state of all components without having them figure it out for themselves -- if the component flag was made "official" this would allow for a community custom loading component to be made. There is no performance impact to this solution.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           name: Install dependencies (dash)
           command: |
               git clone git@github.com:plotly/dash.git
-              git clone git@github.com:plotly/dash-core-components.git
+              git clone -b renderer-regression git@github.com:plotly/dash-core-components.git
               git clone git@github.com:plotly/dash-html-components.git
               git clone git@github.com:plotly/dash-table.git
               . venv/bin/activate
--- a/src/TreeContainer.js
+++ b/src/TreeContainer.js
@@ -75,77 +75,6 @@ class TreeContainer extends Component {
         );
     }
 
-    getLoadingState() {
-        const {
-            _dashprivate_layout: layout,
-            _dashprivate_requestQueue: requestQueue
-        } = this.props;
-
-        const ids = this.isDeepStateElement() ?
-            this.getNestedIds() :
-            (layout && layout.props.id ?
-                [layout.props.id] :
-                []);
-
-        let isLoading = false;
-        let loadingProp;
-        let loadingComponent;
-
-        if (requestQueue) {
-            forEach(r => {
-                const controllerId = isNil(r.controllerId) ? '' : r.controllerId;
-                if (r.status === 'loading' && any(id => contains(id, controllerId), ids)) {
-                    isLoading = true;
-                    [loadingComponent, loadingProp] = r.controllerId.split('.');
-                }
-            }, requestQueue);
-        }
-
-        // Set loading state
-        return {
-            is_loading: isLoading,
-            prop_name: loadingProp,
-            component_name: loadingComponent,
-        };
-    }
-
-    getNestedIds() {
-        const {
-            _dashprivate_layout: layout
-        } = this.props;
-
-        const ids = [];
-        const queue = [layout];
-
-        while (queue.length) {
-            const elementLayout = queue.shift();
-
-            const props = elementLayout &&
-                elementLayout.props;
-
-            if (!props) {
-                continue;
-            }
-
-            const { children, id } = props;
-
-            if (id) {
-                ids.push(id);
-            }
-
-            if (children) {
-                const filteredChildren = filter(
-                    child => !isSimpleComponent(child) && !Registry.resolve(child.type, child.namespace)._dashprivate_deepstate,
-                    Array.isArray(children) ? children : [children]
-                );
-
-                queue.push(...filteredChildren);
-            }
-        }
-
-        return ids;
-    }
-
     getSetProps() {
         return newProps => {
             const {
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           name: Install dependencies (dash)
           command: |
               git clone git@github.com:plotly/dash.git
-              git clone git@github.com:plotly/dash-core-components.git
+              git clone -b renderer-regression git@github.com:plotly/dash-core-components.git
```
Marc-Andre-Rivet: To be removed once in master

src/TreeContainer.js
```diff
@@ -75,28 +75,30 @@ class TreeContainer extends Component {
...
+        const {
+            _dashprivate_layout: layout,
+            _dashprivate_requestQueue: requestQueue
+        } = this.props;
+
+        const ids = this.isDeepStateElement() ?
+            this.getNestedIds() :
+            (layout && layout.props.id ?
+                [layout.props.id] :
+                []);
```
Marc-Andre-Rivet: If component has `deepstate` flag, get all nested ids, otherwise use own id

src/TreeContainer.js
```diff
@@ -107,6 +109,43 @@ class TreeContainer extends Component {
         };
     }
 
+    getNestedIds() {
+        const {
```
Marc-Andre-Rivet: Traverse all children for those with an id. Prune other components with `deepstate` flag set to true

src/TreeContainer.js
```diff
@@ -107,6 +109,43 @@ class TreeContainer extends Component {
...
+
+            if (children) {
+                const filteredChildren = filter(
+                    child => !isSimpleComponent(child) && !Registry.resolve(child.type, child.namespace)._dashprivate_deepstate,
+                    Array.isArray(children) ? children : [children]
+                );
+
+                queue.push(...filteredChildren);
+            }
+        }
```
Marc-Andre-Rivet: Loop in-place on children and append the newly found children until all elements have been tested

src/TreeContainer.js
```diff
@@ -145,7 +184,16 @@ class TreeContainer extends Component {
     }
 
     shouldComponentUpdate(nextProps) {
-        return nextProps._dashprivate_layout !== this.props._dashprivate_layout;
+        const { _dashprivate_layout, _dashprivate_requestQueue } = nextProps;
+
+        return _dashprivate_layout !== this.props._dashprivate_layout ||
+            (this.isDeepStateElement() && _dashprivate_requestQueue !== this.props._dashprivate_requestQueue);
```
Marc-Andre-Rivet: non-deep components only update on layout changes,
deep components update on layout changes and request queue changes

src/TreeContainer.js
```diff
@@ -243,11 +168,78 @@ function mergeProps(stateProps, dispatchProps, ownProps) {
         _dashprivate_dependencies: stateProps.dependencies,
         _dashprivate_dispatch: dispatchProps.dispatch,
         _dashprivate_layout: ownProps._dashprivate_layout,
+        _dashprivate_loadingState: getLoadingState(ownProps._dashprivate_layout, stateProps.requestQueue),
```
Marc-Andre-Rivet: Calculating the loadingState upfront instead of in the render function -- this allows to block/allow re-renderer based on loading state in `shouldComponentUpdate`

src/TreeContainer.js
```diff
@@ -194,12 +165,78 @@ function mergeProps(stateProps, dispatchProps, ownProps) {
...
+
+        const { children, id } = props;
+
+        if (id) {
+            ids.push(id);
+        }
+
+        if (children) {
+            const filteredChildren = filter(
+                child => !isSimpleComponent(child) && !Registry.resolve(child.type, child.namespace)._dashprivate_deepstate,
```
alexcjohnson: `isDeepStateElement`

Marc-Andre-Rivet:
> The one thing I don't understand here is why this gets called _dashprivate_deepstate rather than something more specific like _dashprivate_isLoadingComponent

Just bad naming on my part, I think the idea evolved as I was trying out things and didn't revisit / improved the name afterwards. Renaming to ```_dashprivate_isLoadingComponent``` as suggested.

PR #142, opened 2019-03-29
Dash Clientside Transformations

chriddyp:
This PR enables outputs to be updated clientside with user-defined JavaScript code. It enables users to replace certain Python callbacks with JavaScript callbacks for faster updates, lighter server load, and re-usable components+logic groups.

Unlike the Python callbacks, the JS callback signatures are embedded in the app layout. In this way, it's similar to the approach discussed in https://community.plot.ly/t/could-output-be-moved-inside-the-components-to-improve-readability/17178/2, except done clientside in JavaScript.

Embedding the function signatures in the layout has a couple of main advantages:
1. It more easily enables re-usable code logic+component blocks.
2. It enables slightly easier callbacks with dynamic components (e.g. a TODO list), as the IDs or indices of the dynamic components can be embedded directly in the static argument list.
3. It's relatively easy to read.

This PR introduces two new serializations: the function signatures and the clientside version of `Input`/`State`. These objects are serialized within the initial layout. Here's an example of what this looks like:
```python
{
    "type": "Div",
    "namespace": "dash_html_components",
    "props": {
        "children": [

            {
                "type": "Div",
                "namespace": "dash_html_components",
                "props": {
                    "id": "my-output",
                    "children": {
                        # A simple check for "_dash-type" will indicate to
                        # dash-renderer that this is something "special"
                        # and not user defined.
                        "_dash_type": "function",
                        "function": "my_function",
                        "namespace": "my_library",
                        "positional_arguments": [
                            {
                                "_dash_type": "input",  # or state
                                "id": "my-input",
                                "property": "value"
                            },
                            3  # also allow constants
                        ]
                    }
                }
            },

            {
                "type": "Input",
                "namespace": "dash_core_components",
                "props": {
                    "id": "my-input",
                    "value": "my value"
                }
            }
        ]
    }
}
```

This implementation differs from previous attempts in a few ways:
1. We are no longer prescriptive about what clientside libraries or functions are available to you. You are responsible for writing the functions. Of course, you may find yourself very productive with Ramda 😉 
2. We do not allow updating arbitrarily nested properties of components (e.g. `figure.layout.title`) - only top level properties can be updated. In this way, it mirrors the Python callbacks.
3. The Python interface is light - it just serializes a lookup to the `window[namespace][function]` and descriptions the arguments of that function (which can be a combination of constants & deferred-evaluated Input and State objects)

[...]

PR #143, opened 2019-03-30
Clientside Callbacks

chriddyp:
Next version of https://github.com/plotly/dash-renderer/pull/142, incorporating feedback described in https://github.com/plotly/dash-renderer/pull/142#issuecomment-478160104

Same examples as in #142 will run, however the syntax is a little closer to the python callbacks now. See an associated PR in `dash`

I've pushed up the examples in 5b9934d17851f467321b80eb9bc198e16b678251, I'll remove them before merging this PR. Same GIFs as in #142.


.circleci/config.yml
```diff
@@ -43,6 +43,7 @@ jobs:
               git clone git@github.com:plotly/dash-html-components.git
               git clone git@github.com:plotly/dash-table.git
               . venv/bin/activate
+              cd dash && git checkout clientside-2 && cd ..
```
Marc-Andre-Rivet: Note: You can also do add `-b clientside-2` in the git clone command above.

clientside_0_simple.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@app.callback(
+    Output('output-serverside', 'children'),
+    [Input('input', 'value')])
+def update_output(value):
+    return 'Server says "{}"'.format(value)
+
+
+app.callback(
```
Marc-Andre-Rivet: Code in the Dash review seems misaligned with usage here

- `clientside_function` https://github.com/plotly/dash/pull/672/files#diff-228c9aefac729977642672aa1416bacaR939
- `ClientsideFunction` vs `ClientFunction` https://github.com/plotly/dash/pull/672/files#diff-1b30af980419a25a5ffd5fdfc4ef0173R38

src/actions/index.js
```diff
@@ -458,10 +501,108 @@ function updateOutput(
...
+    // Clientside hook
+    if (!isNil(clientside_function) && !isEmpty(clientside_function)) {
+        let returnValue;
+        try {
+            returnValue = window[clientside_function.namespace][
+                clientside_function.function_name
+            ](
+                ...(has('inputs', payload)
+                    ? pluck('value', payload.inputs)
+                    : []),
```
Marc-Andre-Rivet: For what it's worth, [tested](https://jsperf.com/ramdavanilla) that doing this type of operation with Ramda instead of using a plain js tertianary `payload.inputs ? payload.input.value : []) is about 3 orders of magnitude slower (~1750x). Haven't tested with all usages here but pretty sure this comment will apply to other similar usages as well.

src/actions/index.js
```diff
@@ -371,6 +372,44 @@ function updateOutput(
     const {config, layout, graphs, dependenciesRequest, hooks} = getState();
     const {InputGraph} = graphs;
 
+    const getThisRequestIndex = () => {
```
Marc-Andre-Rivet: Comment not specific to this review but `updateOutput` is in need some deep cleanup. If `TreeContainer` is the rendere's heart, this is certainly the brain. Opening https://github.com/plotly/dash-core/issues/56 for follow up.

src/actions/index.js
```diff
@@ -458,10 +501,108 @@ function updateOutput(
...
+                `The following error occurred while executing ${
+                    clientside_function.namespace
+                }.${clientside_function.function_name} ` +
+                    `in order to update component "${payload.output}" ⋁⋁⋁`
+            );
+            console.error(e);
+            /* eslint-enable no-console */
+
+            /*
+             * Update the request queue by treating a successful clientside
```
Marc-Andre-Rivet: unsuccessful?

src/actions/index.js
```diff
@@ -458,10 +501,108 @@ function updateOutput(
...
+             * like a failed serverside response (500 status code)
+             */
+
+            /* eslint-disable no-magic-numbers */
+            updateRequestQueue(true, 500);
+            /* eslint-enable no-magic-numbers */
+            return;
+        }
+
+        function updateOutput(outputIdAndProp, outputValue) {
```
Marc-Andre-Rivet: Rename. Wrapper function is also called `updateOutput`

tests/test_clientside/clientside.js
```diff
@@ -0,0 +1,27 @@
...
+    add_to_four_outputs: function(value) {
+        return [
+            parseInt(value) + 1,
+            parseInt(value) + 2,
+            parseInt(value) + 3,
+            parseInt(value) + 4
+        ]
+    }
+
+}
```
Marc-Andre-Rivet: As https://github.com/plotly/dash-renderer/pull/143/files#diff-4e3037116dbc2bc69a473b7ace3d02abR508 only supports sync callbacks, would like to see a negative test handling a promise case & failure.

tests/test_render.py
```diff
@@ -2275,3 +2281,265 @@ def set_bc(a):
...
+                    (By.ID, id),
+                    assertions[id]
+                ),
+                message='Failed assertion: #{}.value != {}'.format(
+                    id, assertions[id]
+                )
+            )
+
+
+    def test_simple_clientside_serverside_callback(self):
```
Marc-Andre-Rivet: I would like to see clientside callback tests in their own file

tests/test_render.py
```diff
@@ -2275,3 +2281,265 @@ def set_bc(a):
...
+            ]),
+
+            # clientside
+            html.Label('Mean(x, y, x+y, x+y/2) (clientside)'),
+            dcc.Input(id='mean-of-all-values'),
+
+        ])
+
+        app.clientside_callback(
+            ClientsideFunction('R', 'add'),
```
Marc-Andre-Rivet: As discussed, I think we should isolate clienside callbacks in a predictable dash-property/container in the windows object to prevent arbitrary calls on arbitrary global `window`` object / functions.

src/actions/index.js
```diff
@@ -458,10 +501,108 @@ function updateOutput(
...
+        if (payload.output.startsWith('..')) {
+            /*
+             * If this update is for multiple outputs, then it has
+             * starting & trailing `..` and each propId pair is separated
+             * by `...`, e.g.
+             * "..output-1.value...output-2.value...output-3.value...output-4.value.."
+             */
+            const outputPropIds = payload.output
+                .split('...')
+                .map(o => o.replace('..', ''));
```
Marc-Andre-Rivet: There's now 3 implementations of this multiple-outputs splitting with slight differences., This one, https://github.com/plotly/dash-renderer/blob/master/src/actions/index.js#L247, and https://github.com/plotly/dash-renderer/blob/master/src/reducers/dependencyGraph.js#L26. 🌴

Marc-Andre-Rivet:
Needs an `unreleased` changelog entry

tests/test_render.py
chriddyp: I tried this, but it was compatible with our existing percy config  - https://github.com/plotly/dash-renderer/pull/143/commits/7482e414a7734373a62fd2be5c35b8d9f999339c

tests/test_render.py
chriddyp: Ah, OK I actually don't understand this. So the user wouldn't write `window.their_namespace`, they would write `window.dash_clientside.their_namespace`?

src/actions/index.js
chriddyp: Those tests aren't implementing `pluck` correctly, they are actually implementing `prop` https://ramdajs.com/docs/#prop https://ramdajs.com/docs/#pluck

Marc-Andre-Rivet: I'm not sure I understand. The above uses `pluck` the linked test uses `pluck` not `prop`.
image

tests/test_render.py
Marc-Andre-Rivet: Ok 😢

tests/test_render.py
Marc-Andre-Rivet: Yes. I would suggest that the users JS entrypoint be something like `window.dash_clientside` and ClientsideFunction(namespace, function_name) result in `window.dash_clientside[namespace][function_name]` being called.

src/actions/index.js
```diff
@@ -458,10 +501,108 @@ function updateOutput(
...
+            /* eslint-enable no-console */
+
+            /*
+             * Update the request queue by treating a successful clientside
+             * like a failed serverside response (500 status code)
+             */
+
+            /* eslint-disable no-magic-numbers */
+            updateRequestQueue(true, 500);
+            /* eslint-enable no-magic-numbers */
```
alexcjohnson: I'm not quite clear what `status` gets used for in the request queue, but I'd either put `500` as `STATUS.ERROR` or, perhaps better yet, create a new value just for this purpose so we can tell the difference. We're using `status: 'loading'` in the queue initially, so we're not restricted to http codes... perhaps this can just be `status: 'clientside error'`?

src/actions/index.js
alexcjohnson: Your vanilla code implements `R.prop` and compares it with `R.pluck` called on the wrong data shape (`R.pluck` does something... interesting... when called on an object rather than an array, and maybe that's part of its relative slowness)

It's certainly concise ⛳️ to do this with ramda - I'm curious though, is there actually a case for no `inputs` anymore or is that a relic of `events` days?

src/actions/index.js
```diff
@@ -458,10 +501,108 @@ function updateOutput(
         });
     }
 
+    // Clientside hook
+    if (!isNil(clientside_function) && !isEmpty(clientside_function)) {
```
alexcjohnson: Is there actually a case for an empty `clientside_function` (or any other value that's truthy but missing `namespace` or `function_name`) that we wouldn't want to throw an error? If not I'd just make this `if (clientside_function)` so these other cases fall into the `catch` below rather than silently continuing to the server-side section.

tests/test_render.py
```diff
@@ -2275,3 +2281,265 @@ def set_bc(a):
...
+
+        input = WebDriverWait(self.driver, 10).until(
+            EC.presence_of_element_located((By.ID, "input"))
+        )
+        WebDriverWait(self.driver, 10).until(
+            EC.text_to_be_present_in_element(
+                (By.ID, 'output-serverside'),
+                'Server says "None"'
+            )
+        )
```
alexcjohnson: Can we please add and use the methods we've already got in some of the other repos to reduce some of this repetition?

https://github.com/plotly/dash/blob/87091439ad0a5bb4abf336543e539aa4b0fd5e46/tests/IntegrationTests.py#L24-L33

tests/test_render.py
alexcjohnson: Or we could expose a function like
```js
registerClientside('myNamespace', {
    myFunc: (a, b) => a + b
})
```
That might also be nice later on for validation of function references - though I guess we'd get that from `window.dash_clientside` as well - but with a registration function it would be easier to ensure that multiple calls wouldn't clobber each other (ie the user writes something like `window.dash_clientside = {namespace: {...} }` in two different files)

alexcjohnson: It is nice allowing access to ramda out of the box (both as a clientside namespace on its own and off `window` so you can use it from your own clientside functions)... how did it get to be `window.R` though? I feel like it'd be better as `ramda`, at least for its clientside namespace, especially with dashR coming up.

src/actions/index.js
Marc-Andre-Rivet: I see. You're right, I'm comparing apples and oranges. The Ramda performance is more than fine when testing the right thing.

chriddyp: >  I'm curious though, is there actually a case for no inputs anymore or is that a relic of events days?

there is not. good catch!

src/actions/index.js
chriddyp: 🙈

tests/test_render.py
chriddyp: > how did it get to be window.R though? I feel like it'd be better as ramda, at least for its clientside namespace, especially with dashR coming up.

it's actually not here out of the box! I included the ramda bundle from their CDN and put it in the test `assets/`

chriddyp: I think I'll start with just regular namespaces, then do register in a follow up PR

src/actions/index.js
chriddyp: 🙈

tests/test_render.py
chriddyp: 😸

chriddyp: but these functions get marginally better every time i write a test....

chriddyp: just kidding - 🔪 'ing away

tests/test_render.py
alexcjohnson: > it's actually not here out of the box! I included the ramda bundle from their CDN and put it in the test assets/

Ah I see, `window.R` is how the CDN version exports itself... OK, as long as that's not added by dash itself that's fine.

> I think I'll start with just regular namespaces, then do register in a follow up PR

Sure, the boilerplate you need right now is a bit steep but it should be easy to add a register fn later in a backward-compatible way.

chriddyp:
Alright, thanks everyone for your review. I believe I have addressed y'alls comments now.

tests/test_render.py
chriddyp: actually lol I spoke too soon. I had to modify the implementaiton under the hood a bit to handle `<input/>` `value`. So, it is truly a tiny bit better now. same API though!
Here's the commit - https://github.com/plotly/dash-renderer/pull/143/commits/374493df3403e7a5da55a73016f348d62d6dc2a3

alexcjohnson: Ha, I didn't even see those ones that you modified - I guess sometimes we have these in the test file, sometimes in the base class... anyway yeah, nice tweak! (cc @byronz for whenever we manage to share these)

.circleci/config.yml
alexcjohnson: non-⛔️ as this line will disappear when the companion PR merges... but we may benefit 🐎 from using a shallow clone `git clone -depth=1` on CI, in which case `-b <branch>` may be the required method.

PR #150, opened 2019-04-11
Issue148 - `0` children regression in dash==0.40.0

### LINKED ISSUE #148: Regression going from Dash 0.39 to 0.40 -- implicit `0` child is missing from app
Note the difference in the last table.

Code
```
import dash
import pandas as pd

df = pd.read_csv(
    'https://gist.githubusercontent.com/chriddyp/'
    'c78bf172206ce24f77d6363a2d754b59/raw/'
    'c353e8ef842413cae56ae3920b8fd78468aa4cb2/'
    'usa-agricultural-exports-2011.csv')


def generate_table(dataframe, max_rows=10):
    return html.Div([
        html.Table(
            # Header
            [html.Tr([html.Th(col) for col in dataframe.columns])] +

            # Body
            [html.Tr([
                html.Td(children=[dataframe.iloc[i][col]]) for col in dataframe.columns
            ]) for i in range(min(len(dataframe), max_rows))]
        ),
        html.Table(
            # Header
            [html.Tr([html.Th(col) for col in dataframe.columns])] +

            # Body
            [html.Tr([
                html.Td([dataframe.iloc[i][col]]) for col in dataframe.columns
            ]) for i in range(min(len(dataframe), max_rows))]
        ),
        html.Table(
            # Header
            [html.Tr([html.Th(col) for col in dataframe.columns])] +

            # Body
            [html.Tr([
                html.Td(dataframe.iloc[i][col]) for col in dataframe.columns
            ]) for i in range(min(len(dataframe), max_rows))]
        )
    ])

external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']

app = dash.Dash(__name__, external_stylesheets=external_stylesheets)

app.layout = html.Div(children=[
    html.H4(children='US Agriculture Exports (2011)'),
    generate_table(df, 5)
])

if __name__ == "__main__":
    app.run_server(port=8070)
```

Dash 0.39
image

Dash 0.40
image

Marc-Andre-Rivet:
FIxes #148.

Found the issue in dash-docs percy deltas in https://percy.io/plotly/dash-docs/builds/1719186
image

Add test for array case `children=[0]`, this passes with the existing code: https://github.com/plotly/dash-renderer/commit/0602a155e12ba718b1a9bc764b93e32c6bd78004

Add test for non-array case `children=0`, this fails with the existing code: https://github.com/plotly/dash-renderer/commit/d48e81dd2f52434a7fb86a707a8483e26e779078

Fix: https://github.com/plotly/dash-renderer/commit/b600b8cd296dd0ade14737dc6985e4596b10e736

### DIFF
--- a/src/TreeContainer.js
+++ b/src/TreeContainer.js
@@ -33,7 +33,7 @@ const createContainer = component => isSimpleComponent(component) ?
 
 class TreeContainer extends Component {
     getChildren(components) {
-        if (!components) {
+        if (isNil(components)) {
             return null;
         }
 
--- a/tests/test_render.py
+++ b/tests/test_render.py
@@ -104,7 +104,7 @@ class Tests(IntegrationTests):
 
         if expected_length is not None:
             self.assertEqual(len(request_queue), expected_length)
-    
+
     def test_initial_state(self):
         app = Dash(__name__)
         app.layout = html.Div([
@@ -472,6 +472,19 @@ class Tests(IntegrationTests):
         self.request_queue_assertions(0)
 
         self.percy_snapshot(name='layout')
+
+        assert_clean_console(self)
+
+    def test_nully_child(self):
+        app = Dash(__name__)
+        app.layout = html.Div(id='nully-wrapper', children=[0])
+
+        self.startServer(app)
+
+        self.wait_for_element_by_css_selector('#nully-wrapper')
+        wrapper = self.driver.find_element_by_id('nully-wrapper')
+
+        self.assertEqual(wrapper.get_attribute('innerHTML'), '0')
 
         assert_clean_console(self)
 
--- a/tests/test_render.py
+++ b/tests/test_render.py
@@ -481,10 +481,7 @@ class Tests(IntegrationTests):
 
         self.startServer(app)
 
-        self.wait_for_element_by_css_selector('#nully-wrapper')
-        wrapper = self.driver.find_element_by_id('nully-wrapper')
-
-        self.assertEqual(wrapper.get_attribute('innerHTML'), '0')
+        self.wait_for_text_to_equal('#nully-wrapper', '0')
 
         assert_clean_console(self)
 
@@ -494,10 +491,7 @@ class Tests(IntegrationTests):
 
         self.startServer(app)
 
-        self.wait_for_element_by_css_selector('#nully-wrapper')
-        wrapper = self.driver.find_element_by_id('nully-wrapper')
-
-        self.assertEqual(wrapper.get_attribute('innerHTML'), '0')
+        self.wait_for_text_to_equal('#nully-wrapper', '0')
 
         assert_clean_console(self)
 
--- a/tests/test_render.py
+++ b/tests/test_render.py
@@ -475,7 +475,7 @@ class Tests(IntegrationTests):
 
         assert_clean_console(self)
 
-    def test_array_of_nully_child(self):
+    def test_array_of_falsy_child(self):
         app = Dash(__name__)
         app.layout = html.Div(id='nully-wrapper', children=[0])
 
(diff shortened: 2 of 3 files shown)


src/TreeContainer.js
```diff
@@ -33,7 +33,7 @@ const createContainer = component => isSimpleComponent(component) ?
 
 class TreeContainer extends Component {
     getChildren(components) {
-        if (!components) {
+        if (isNil(components)) {
```
Marc-Andre-Rivet: 0 is falsy..

tests/test_render.py
```diff
@@ -475,6 +475,32 @@ def test_initial_state(self):
 
         assert_clean_console(self)
 
+    def test_array_of_nully_child(self):
+        app = Dash(__name__)
+        app.layout = html.Div(id='nully-wrapper', children=[0])
```
Marc-Andre-Rivet: This worked and needs to keep working

tests/test_render.py
```diff
@@ -475,6 +475,32 @@ def test_initial_state(self):
...
+        self.wait_for_element_by_css_selector('#nully-wrapper')
+        wrapper = self.driver.find_element_by_id('nully-wrapper')
+
+        self.assertEqual(wrapper.get_attribute('innerHTML'), '0')
+
+        assert_clean_console(self)
+
+    def test_of_nully_child(self):
+        app = Dash(__name__)
+        app.layout = html.Div(id='nully-wrapper', children=0)
```
Marc-Andre-Rivet: This is the regression case

tests/test_render.py
```diff
@@ -104,7 +104,7 @@ def request_queue_assertions(
 
         if expected_length is not None:
             self.assertEqual(len(request_queue), expected_length)
-
```
alexcjohnson: Oh noes, trailing whitespace courtesy of @chriddyp ??? 😅

chriddyp: no way!! 🙊 🙊 🙊

tests/test_render.py
alexcjohnson: nice. I'd call this falsy though, not nully.

tests/test_render.py
```diff
@@ -475,6 +475,32 @@ def test_initial_state(self):
...
+    def test_array_of_nully_child(self):
+        app = Dash(__name__)
+        app.layout = html.Div(id='nully-wrapper', children=[0])
+
+        self.startServer(app)
+
+        self.wait_for_element_by_css_selector('#nully-wrapper')
+        wrapper = self.driver.find_element_by_id('nully-wrapper')
+
+        self.assertEqual(wrapper.get_attribute('innerHTML'), '0')
```
alexcjohnson: Can't this whole section just be [`self.wait_for_text_to_equal('#nully-wrapper', '0')`](https://github.com/plotly/dash-renderer/blob/master/tests/test_render.py#L59)?

tests/test_render.py
Marc-Andre-Rivet: good point

tests/test_render.py
Marc-Andre-Rivet: Hum.. wasn't aware of that one. Looks like `wait_for_text_to_equal` uses `value` and that would mean it's only meant to be used with `<input />` elements.

chriddyp: `value` _or_ `.text`:
https://github.com/plotly/dash-renderer/blob/2c0169c52f5d604db48285814836fe394a0aea07/tests/test_render.py#L63

Marc-Andre-Rivet: I can't read code it seems. Will update to use suggested.

tests/test_render.py
byronz: too late for the comment, but this should be merged with the `test_initial_state`.

PR #152, opened 2019-04-11
more devtools enhancements

chriddyp:
mostly css but a few bug fixes as well

image

### DIFF
--- a/src/components/error/FrontEnd/FrontEndError.react.js
+++ b/src/components/error/FrontEnd/FrontEndError.react.js
@@ -34,7 +34,6 @@ class FrontEndError extends Component {
                 onClick={() => this.setState({collapsed: !collapsed})}
             >
                 <span className="dash-fe-error-top__group">
-                    ☣️&nbsp;
                     <span className="dash-fe-error__title">
                         {e.error.message || 'Error'}
                     </span>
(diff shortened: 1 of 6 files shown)


src/components/error/FrontEnd/FrontEndError.react.js
```diff
@@ -34,7 +34,6 @@ class FrontEndError extends Component {
                 onClick={() => this.setState({collapsed: !collapsed})}
             >
                 <span className="dash-fe-error-top__group">
-                    ☣️&nbsp;
```
byronz: image

byronz:
@chriddyp  can you share your sample app too?

chriddyp:
```python
import dash

import datetime

app = dash.Dash(__name__)
app.css.config.serve_locally = True


app.layout = html.Div([
    html.Button(id='python', children='Python exception', n_clicks=0),
    html.Button(id='validation', children='Validation error', n_clicks=0),
    html.Button(id='normal', children='Regular execution', n_clicks=0),
    html.Div(id='output')
])

@app.callback(
    Output('output', 'children'),
    [
        Input('python', 'n_clicks'),
        Input('validation', 'n_clicks'),
        Input('normal', 'n_clicks'),
    ])
def update_output(*args):
    if all([a > 0 for a in args]):
        return ''

    triggered = [t['prop_id'] for t in dash.callback_context.triggered]
    if 'python.n_clicks' in triggered:
        1/0
    elif 'validation.n_clicks' in triggered:
        return dcc.Graph(animate=0)
    else:
        return 'Callback returned at {}'.format(datetime.datetime.now())


if __name__ == '__main__':
    app.run_server(debug=True, dev_tools_hot_reload=False)
```

PR #153, opened 2019-04-12
another round of dev-tools edits

chriddyp:
- explicitly handle the nested array case (`html.Div(children=[['oh no']])`)
- explicitly handle dicts as children: (`html.Div(children={'why': 'oh why'}`)
- a few css edits

### DIFF
--- a/src/TreeContainer.js
+++ b/src/TreeContainer.js
@@ -37,9 +37,9 @@ function validateComponent(componentDefinition) {
         );
     }
     if (type(componentDefinition) === 'Object' &&
-            !has('namespace', componentDefinition) &&
-            !has('type', componentDefinition) &&
-            !has('props', componentDefinition)) {
+            !(has('namespace', componentDefinition) &&
+              has('type', componentDefinition) &&
+              has('props', componentDefinition))) {
         throw new Error(
             'An object was provided as `children` instead of a component, ' +
             'string, or number (or list of those). ' +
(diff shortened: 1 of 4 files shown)


chriddyp:
dict case
image

nested lists
image

chriddyp:
and here are the test cases that i'm working from: https://gist.github.com/chriddyp/93d5c12b88d7333f56457b867e94d7b6

src/TreeContainer.js
```diff
@@ -25,6 +26,29 @@ import { assertPropTypes } from 'check-prop-types';
...
+            'of just a list. ' +
+            'Check the component that has the following contents, ' +
+            'and remove of the levels of nesting: \n' +
+            JSON.stringify(componentDefinition, null, 2)
+        );
+    }
+    if (type(componentDefinition) === 'Object' &&
+            !has('namespace', componentDefinition) &&
+            !has('type', componentDefinition) &&
+            !has('props', componentDefinition)) {
```
byronz: I feel like the last two `&&` should be `||`

src/components/error/GlobalErrorOverlay.react.js
```diff
@@ -14,14 +15,10 @@ export default class GlobalErrorOverlay extends Component {
...
-                errors = error.frontEnd;
-            }
-
-            error.backEnd.forEach(backEndError => {
-                errors.push(backEndError);
-            });
+            const errors = concat(
+                error.frontEnd,
+                error.backEnd
+            );
```
chriddyp: @byronz - this should fix 
image

byronz: good catch ! :+1:

PR #155, opened 2019-04-13
another round of dev tools edits

chriddyp:
Improving the prop type error messages.

Previously
image

***

Now
image

***

image

***

image


chriddyp:
The second set of commits simplifies the error overlays. See 9010044 for details.
In the previous version, we:
1. wrapped the output component with an overlay. This could cause flow
issues as the overlay wrapper was a div with `display: block` which doesn't necessarily match the wrapped component. For example, the output component could be a `span` (`display: inline`) in-line with other spans.
2. Disabled the input components

This commit simplifies the logic: now we just display the previous state of the component (if available) without an overlay and without disabling the inputs.

This is similar to how the backend errors affect the frontend.

error-types

chriddyp:
Now that we have nice error messages, the stack traces aren't as useful for the end user. I feel like they may even indicate to the user that the issue _wasn't their fault_ but was something much more complicated. 
image

byronz:
@chriddyp  python3.x test errors are different than 2.7 
```
ERROR: test_devtools_validation_errors (tests.test_render.Tests)
  File "/home/circleci/repo/tests/test_render.py", line 2647, in test_devtools_validation_errors
    self.wait_for_element_by_css_selector('.test-devtools-error-toggle').click()
  File "/home/circleci/repo/tests/test_render.py", line 60, in wait_for_element_by_css_selector
    'Could not find element with selector "{}"'.format(selector)
  File "/home/circleci/repo/venv/lib/python3.7/site-packages/selenium/webdriver/support/wait.py", line 80, in until
    raise TimeoutException(message, screen, stacktrace)
selenium.common.exceptions.TimeoutException: Message: Could not find element with selector ".test-devtools-error-toggle"
```

chriddyp:
alright - those tests have been fixed and several more have been added

src/TreeContainer.js
```diff
@@ -57,16 +58,26 @@ const createContainer = component => isSimpleComponent(component) ?
...
-        'component prop', element);
+    const {
+        element,
+        extraProps,
+        props,
+        children,
+        type
+    } = p;
+
+    try {
```
byronz: if you want to parse the error result, better use `checkPropTypes`  https://www.npmjs.com/package/check-prop-types#usage
`assertPropTypes` is a helper function to throw an error directly from the result.

src/components/error/ComponentErrorBoundary.react.js
```diff
@@ -126,16 +62,13 @@ class UnconnectedComponentErrorBoundary extends Component {
 UnconnectedComponentErrorBoundary.propTypes = {
     children: PropTypes.object,
     componentId: PropTypes.string,
-    componentType: PropTypes.string,
-    dispatch: PropTypes.func,
     error: PropTypes.object,
-    graphs: PropTypes.object,
+    dispatch: PropTypes.function,
```
byronz: typo? PropTypes.func,

src/exceptions.js
```diff
@@ -0,0 +1,119 @@
+import {contains, has} from 'ramda';
+
+export function propTypeErrorHandler(e, props, type) {
+    /*
+     * propType error messages are constructed in
+     * https://github.com/facebook/prop-types/blob/v15.7.2/factoryWithTypeCheckers.js
```
byronz: interesting, it looks like after this merge, https://github.com/facebook/prop-types/pull/87/files  no one was updating the documentation at all, so unless you dig into the PR storybook, it's a hidden ruby. or is it possible that the strict way is not recommended?

src/TreeContainer.js
chriddyp: Yeah good point. Something like this?
```

    const msg = checkPropTypes(element.propTypes, props, 'component prop', element);
    if (msg !== undefined) {
        propTypeErrorHandler(e, props, type);
    }
```

chriddyp: I think I find the `try` / `catch` flow easier to understand as is, but I can change if you'd like 😸

src/exceptions.js
chriddyp: Yeah I can't quite figure it out. I get the impression that the team maintaining `prop-types` is pretty overloaded. I think that most of the React team itself has moved over to flow/type script and so doesn't depend on prop-types as much anymore. Not sure though.

src/TreeContainer.js
byronz: I prefer to change as we were doing throw and catch game => 自娱自乐, 
let's save some CO2 emission here :)

byronz:
💃  avant dodo

PR #156, opened 2019-04-14
dev tools screenshot tests

chriddyp:
~(work in progress)~ pretty happy with how these came out


chriddyp:
Dang, unable to find the devtools UI window on CircleCI. Not sure what's going wrong here.
```
ERROR: test_devtools_python_errors (tests.test_render.Tests)
  File "/home/circleci/repo/tests/test_render.py", line 2289, in test_devtools_python_errors
    self.wait_for_element_by_css_selector('.test-devtools-error-toggle').click()
  File "/home/circleci/repo/tests/test_render.py", line 60, in wait_for_element_by_css_selector
    'Could not find element with selector "{}"'.format(selector)
  File "/home/circleci/repo/venv/lib/python3.6/site-packages/selenium/webdriver/support/wait.py", line 80, in until
    raise TimeoutException(message, screen, stacktrace)
selenium.common.exceptions.TimeoutException: Message: Could not find element with selector ".test-devtools-error-toggle"
```

This does work locally.

chriddyp:
Hm, devtools UI is there but the error messages aren't popping up
image

chriddyp:
Perhaps it's an issue with the build systems / branches, like some other uncaught exception. I'll try a fresh build tomorrow.

chriddyp:
Re-built everything locally using same instructions in `circle/config.yml` and was able to reproduce the error locally 👍

chriddyp:
Selenium screenshot artifact looks (mostly) good now:
image

chriddyp:
hooray we have percy snapshots 🎉 
image

I suspect the missing CSS is from werkzeug's JS - it adds some dynamic CSS to the nodes. It might be doing that in a way that is un-capturable (i.e. in memory)

tests/IntegrationTests.py
```diff
@@ -21,6 +21,7 @@ def percy_snapshot(cls, name=''):
         cls.percy_runner.snapshot(
             name=snapshot_name
         )
+        cls.driver.save_screenshot('/tmp/artifacts/{}.png'.format(name))
```
byronz: If I understood the purpose correctly, this is to investigate with circile ci docker? once shift to pytest, we can implement a hook to do this automatically when the test case failed.

tests/test_render.py
```diff
@@ -2257,3 +2258,324 @@ def test_clientside_fails_when_returning_a_promise(self):
...
+
+        self.wait_for_element_by_css_selector('#python').click()
+        self.percy_snapshot('devtools - python exception - closed')
+        self.wait_for_element_by_css_selector('.test-devtools-error-toggle').click()
+        self.percy_snapshot('devtools - python exception - open')
+
+        self.wait_for_element_by_css_selector('#python').click()
+        self.percy_snapshot('devtools - python exception - 2 errors')
+        self.wait_for_element_by_css_selector('.test-devtools-error-toggle').click()
+        self.percy_snapshot('devtools - python exception - 2 errors open')
```
byronz: @chriddyp I'm afraid some of the percy snapshots will need continuously approval like: time stamp, library change the line number etc.

image

tests/test_render.py
```diff
@@ -2257,3 +2258,324 @@ def test_clientside_fails_when_returning_a_promise(self):
...
+
+    def test_devtools_validation_errors_in_place(self):
+        app = dash.Dash(__name__)
+
+        app.layout = html.Div([
+            html.Button(id='button', children='update-graph', n_clicks=0),
+            dcc.Graph(id='output', figure={'data': [{'y': [3, 1, 2]}]})
+        ])
+
+        # animate is a bool property
```
byronz: a good example of how to hide the intent if without the comment :thought_balloon:

tests/IntegrationTests.py
chriddyp: yeah exactly. we can remove this now that the issue has been fixed. Doing it automatically when tests fail would be great though!

tests/test_render.py
chriddyp: ah yeah :( I For the timestamp, I think I'll give https://docs.percy.io/docs/percy-specific-css a try. For the line numbers, I think they'll change less frequently since the exception remains the same. We could even pin `flask` too.

chriddyp: Looking back at this - for the line numbers, it looks like the `test_render` traceback doesn't appear in this screenshot test (its hidden behind the scroll). So, for now I don't think it'll be an issue but it could be in the future if those exceptions become visible.

src/components/error/FrontEnd/FrontEndError.react.js
```diff
@@ -40,9 +41,12 @@ class FrontEndError extends Component {
                     </span>
                 </span>
                 <span className="dash-fe-error-top__group">
-                    <span className="dash-fe-error__timestamp">
+                    <span className="dash-fe-error__timestamp percy-hide">
                         {`${e.timestamp.toLocaleTimeString()}`}
                     </span>
+                    <span className="dash-fe-error__timestamp percy-show">
+                        {'12:49:14 PM'}
+                    </span>
```
chriddyp: 😸 
image

src/components/error/FrontEnd/FrontEndError.react.js
```diff
@@ -40,9 +41,12 @@ class FrontEndError extends Component {
                     </span>
                 </span>
                 <span className="dash-fe-error-top__group">
-                    <span className="dash-fe-error__timestamp">
+                    <span className="dash-fe-error__timestamp percy-hide">
                         {`${e.timestamp.toLocaleTimeString()}`}
                     </span>
+                    <span className="dash-fe-error__timestamp percy-show">
+                        {'12:49:14 PM'}
```
byronz: :cow2:  maybe a special hardcode time like 00:00:00 to remind this is a special hack of percy?

src/components/error/Percy.css
```diff
@@ -0,0 +1,12 @@
+.percy-show {
+    display: none;
+}
+
+@media only percy {
```
chriddyp: https://docs.percy.io/docs/percy-specific-css

src/components/error/FrontEnd/FrontEndError.react.js
chriddyp: Hide the time-varying timestamps on percy and display a constant timestamp so that we can still see what the UI looks like.

tests/test_render.py
chriddyp: Timestamp fixed in https://github.com/plotly/dash-renderer/pull/156/commits/ec81d67ebdc33ce1e216795962c25efeb0939665

src/components/error/FrontEnd/FrontEndError.react.js
chriddyp: good idea. I'll keep the `PM` so that the width is the same

byronz:
:dancer:  nice solution with percy swapping, 
image

PR #157, opened 2019-04-14
Fix hot reload

chriddyp:
Fixes https://github.com/plotly/dash/issues/634

### DIFF
--- a/src/components/core/Reloader.react.js
+++ b/src/components/core/Reloader.react.js
@@ -11,7 +11,6 @@ class Reloader extends React.Component {
         if (props.config.hot_reload) {
             const {interval, max_retry} = props.config.hot_reload;
             this.state = {
-                hash: null,
                 interval,
                 disabled: false,
                 intervalId: null,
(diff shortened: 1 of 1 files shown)


chriddyp:
hotreload

chriddyp:
It looks like our existing hot reload tests don't test python changes yet (only css changes).

I think that getting proper python integration tests here might be a bit more work, so I'd like to get this fix merged in first and then loop back later on integration tests.

chriddyp:
Best to review this one commit-by-commit

src/components/core/Reloader.react.js
```diff
@@ -90,29 +125,23 @@ class Reloader extends React.Component {
...
-                    window.clearInterval(this.state.intervalId);
+                    // Backend code changed - can do a soft reload in place
                     dispatch({type: 'RELOAD'});
                 }
             }
         } else if (reloadRequest.status === 500) {
             if (this._retry > this.state.max_retry) {
-                window.clearInterval(this.state.intervalId);
+                this.clearInterval();
                 // Integrate with dev tools ui?!
```
byronz: was wondering if we should throw and the error and see the comment. yes!

src/components/core/Reloader.react.js
```diff
@@ -124,16 +174,20 @@ class Reloader extends React.Component {
         const {dispatch} = this.props;
         const {disabled, interval} = this.state;
         if (!disabled && !this.state.intervalId) {
-            const intervalId = setInterval(() => {
-                dispatch(getReloadHash());
+            const intervalId = window.setInterval(() => {
+                // Prevent requests from piling up - reloading can take
```
byronz: :cat:

byronz:
my current understanding of this part leading to a black box reviewing, I verified the fix with the reloading issue I see with dash-docs, from the network traffic, it seems works properly now with both hard and soft reload.  :dancer:  

better have another approval from @alexcjohnson  or @Marc-Andre-Rivet

PR #158, opened 2019-04-15
styling chainvis

nicolaskruchten:
Here's what the new styling looks like:

image


chriddyp:
Maybe `overflow: auto` so the scrollbars don't appear if they aren't necessary? Otherwise 💃

PR #159, opened 2019-04-17
:recycle: use safe way

byronz:
feel like breaking your private list, hesitate for 5s about `_dashprivate_config`  :no_bell:

### DIFF
--- a/src/TreeContainer.js
+++ b/src/TreeContainer.js
@@ -98,6 +98,8 @@ class TreeContainer extends Component {
     }
 
     getComponent(_dashprivate_layout, children, loading_state, setProps) {
+        const { config } = this.props;
+
         if (isEmpty(_dashprivate_layout)) {
             return null;
         }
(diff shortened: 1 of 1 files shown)


chriddyp:
:+1:

src/TreeContainer.js
```diff
@@ -294,6 +298,7 @@ export const AugmentedTreeContainer = connect(
         _dashprivate_loadingState: getLoadingState(ownProps._dashprivate_layout, stateProps.requestQueue),
         _dashprivate_paths: stateProps.paths,
         _dashprivate_requestQueue: stateProps.requestQueue,
+        config: stateProps.config,
```
Marc-Andre-Rivet: Please rename as `_dashprivate_config` -- since these props are public / accessible from custom components, and yet purely internal details, the only protection we have from people using them (and they have) is by convention, prepending the name so it's obvious they are doing so at their own risk.

PR #165, opened 2019-04-19
Prerelease QA branch

chriddyp:
- [x] fix height of DAG on smaller screens

PR #166, opened 2019-04-19
Fix setProps when IDs aren't provided

chriddyp:
Fixes #165

### DIFF
--- a/src/TreeContainer.js
+++ b/src/TreeContainer.js
@@ -4,7 +4,9 @@ import Registry from './registry';
 import {propTypeErrorHandler} from './exceptions';
 import {connect} from 'react-redux';
 import {
+    addIndex,
     any,
+    concat,
     contains,
     filter,
     forEach,
@@ -15,6 +17,7 @@ import {
     map,
     mergeAll,
     omit,
+    path,
     pick,
     propOr,
     type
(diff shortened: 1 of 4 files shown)


src/TreeContainer.js
```diff
@@ -87,14 +90,16 @@ CheckedComponent.propTypes = {
...
             return null;
         }
 
         return Array.isArray(components) ?
-            map(createContainer, components) :
-            createContainer(components);
+            addIndex(map)(
+                (component, i) => createContainer(component, concat(path, ['props', 'children', i])),
+                components
+            ) : createContainer(components, concat(path, ['props', 'children']));
```
alexcjohnson: Non-🚫 but to my eye ramda makes this much harder to read, vs vanilla
```
components.map((component, i) =>
    createContainer(component, concat(path, ['props', 'children', i]))
)
```

tests/test_render.py
```diff
@@ -2960,3 +2961,27 @@ def display_content(pathname):
...
+            app,
+            debug=True,
+            use_reloader=False,
+            use_debugger=True,
+            dev_tools_hot_reload=False,
+        )
+
+        self.wait_for_element_by_css_selector('#id').send_keys('hello input with ID')
+
+        self.wait_for_element_by_css_selector('#container input').send_keys('hello input without ID')
```
alexcjohnson: Just because of the issues we've had with percy autoapproval, would you mind testing the desired outcome in code here?

chriddyp: Yeah good idea - easy enough to do here anyway.

alexcjohnson:
@chriddyp this is awesome (my :dancer: stands BTW, despite the extra discussion). But @byronz merged some of the `if(id)` hacks in https://github.com/plotly/dash-core-components/pull/523, do we need to remove them now? Seems like they could actually be harmful once this PR is merged.

chriddyp:
Yeah - I think I'll follow up with a PR that cleans up some of the DCC logic. I think that the `id => setState` is actually still "safe" but it could be cleaned up - many of the components can become functional & stateless

PR #170, opened 2019-05-09
204 NO CONTENT regression fix

Marc-Andre-Rivet:
Regression between Dash 0.41 and 0.42

Reproducible with Byron's code below:
```
import dash
from multiprocessing import Value
from dash import callback_context, no_update
from dash.exceptions import PreventUpdate
from multiprocessing import Value

initial_input = 'initial input'
initial_output = 'initial output'


app = dash.Dash(__name__)
# app.config['suppress_callback_exceptions'] = True
app.css.config.serve_locally = True

app.layout = html.Div([
    dcc.Input(id='input', value=initial_input),
    html.Div(initial_output, id='output1'),
    html.Div(initial_output, id='output2'),
])

callback1_count = Value('i', 0)
callback2_count = Value('i', 0)


@app.callback(Output('output1', 'children'), [Input('input', 'value')])
def callback1(value):
    callback1_count.value += 1
    if callback1_count.value > 2:
        return no_update
    raise PreventUpdate("testing callback does not update")


@app.callback(Output('output2', 'children'), [Input('output1', 'children')])
def callback2(value):
    callback2_count.value += 1
    return value

if __name__ == '__main__':
    app.run_server(debug=True)
```

Introduced in: https://github.com/plotly/dash-renderer/pull/152/files#diff-4e3037116dbc2bc69a473b7ace3d02abR678 and https://github.com/plotly/dash-renderer/pull/162/files#diff-4e3037116dbc2bc69a473b7ace3d02abR682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+### Fixed
+- Fix regression on handling PreventUpdate (204 NO CONTENT) [#170](https://github.com/plotly/dash-renderer/pull/170)
+
 ## [0.23.0] - 2019-04-25
 ### Fixed
 - Fix regression for `children=0` case [#148](https://github.com/plotly/dash-renderer/issues/148)
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -680,6 +680,11 @@ function updateOutput(
                  */
                 if (res.status !== STATUS.PREVENT_UPDATE) {
                     throw res;
+                } else {
+                    /*
+                     * This is a 204 response code, there's no content to process.
+                     */
+                    return;
                 }
             }
 
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -675,16 +675,16 @@ function updateOutput(
                 updateRequestQueue(true, res.status);
 
                 /*
-                * This is a 204 response code, there's no content to process.
-                */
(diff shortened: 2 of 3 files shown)


byronz:
> But I do think this warrants a test case.

the above sample app from one of integration would be able to catch the issue
but failed to notify as I believe there is a bug in old implementation of check clear console log in selenium. consolidate the IntegrationTest for both dash and dash-renderer with `desired_capabilities` disclosed it.  yes but we would need more customized case later

alexcjohnson:
> consolidate the IntegrationTest for both dash and dash-renderer with desired_capabilities disclosed it.

Ok, if you have a test to :lock: this elsewhere, we can merge this PR without one. Though it looks like there’s a test already that’s sensitive to this and needs fixing?

Marc-Andre-Rivet:
Not sure what's going on here but the error was already present when the renderer was bumped to `0.23`: https://circleci.com/gh/plotly/dash-renderer/2189?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Marc-Andre-Rivet:
Ah! So this comes from this change in DCC: https://github.com/plotly/dash-core-components/blame/master/src/components/Input.react.js#L103

As previously discussed, some attention to the builds is still required when bumping versions as some combinations of packages don't get automatically tested yet. In this case dcc was last updated after renderer and the diff was only caught at release.

Updating the test.

Marc-Andre-Rivet:
The percy diffs above are caused by the previous failure in master erasing the baseline screenshots for  `test_callbacks_generating_children`

PR #172, opened 2019-05-13
[WIP] Maintenance - do not merge

alexcjohnson:
Cleanup melange. @byronz this ended up running to 15 files... will be a little annoying to merge into the monorepo, my apologies.


package.json
```diff
@@ -37,7 +37,7 @@
     "prop-types": "^15.6.2",
     "query-string": "^4.3.2",
     "radium": "^0.22.1",
-    "ramda": "^0.23.0",
+    "ramda": "^0.26.1",
```
alexcjohnson: I wanted `mergeDeepLeft` so I upgraded `ramda` - but there were two breaking changes there: changed order of args in `adjust`, and removed `import R from 'ramda'` - you can still do `import * as R` but seems like we're moving to explicit imports anyway.

tests/test_render.py
```diff
@@ -1876,6 +1875,11 @@ def test_hot_reload(self):
                 '#hot-reload-content', 'background-color', 'rgba(255, 0, 0, 1)'
             )
         finally:
+            # load a blank page, so we don't get errant _reload_hash
+            # network failures
+            self.driver.get('about:blank')
```
alexcjohnson: `test_initial_state` is next (alphabetically) after `test_hot_reload`, and its console was filled up with
```
http://localhost:8050/_reload-hash - Failed to load resource: net::ERR_CONNECTION_REFUSED
```
because the hot reloading server had been stopped, and I guess our `clear_log` call isn't happening at the right time in the sequence to wipe these.

alexcjohnson:
@byronz I'll take care of bringing the still-relevant pieces of this over to the `dash` repo after the merge. Meantime I'll close this PR

PR #174, opened 2019-05-14
Build & slew of minor improvements

Marc-Andre-Rivet:
Similar to:
https://github.com/plotly/dash-core-components/pull/555
https://github.com/plotly/dash-html-components/pull/113

Tagging @rpkyle for potential impact on R, and @byron for potential impact on mono repo.


.npmignore
```diff
@@ -1,8 +0,0 @@
-node_modules/
-.npm
-.git/
-vv/
-venv/
-*.pyc
-*.log
-.DS_Store
```
alexcjohnson: Are we not publishing this to npm anymore? Or some other reason this is unused?

alexcjohnson:
As discussed with @Marc-Andre-Rivet we'll revisit this in the combined repo rather than trying to figure out what's going wrong here.

PR #175, opened 2019-05-20
Hide undo redo

alexcjohnson:
Companion to https://github.com/plotly/dash/pull/724


tests/test_render.py
```diff
@@ -154,11 +154,12 @@ def test_initial_state(self):
         self.startServer(app)
         el = self.wait_for_element_by_css_selector('#react-entry-point')
 
+        # Note: this .html file shows there's no undo/redo button by default
         _dash_app_content_html = os.path.join(
             os.path.dirname(__file__),
             'test_assets', 'initial_state_dash_app_content.html')
         with open(_dash_app_content_html) as fp:
-            rendered_dom = BeautifulSoup(fp.read(), 'lxml')
+            rendered_dom = BeautifulSoup(fp.read().strip(), 'lxml')
```
alexcjohnson: `strip` because my editor doesn't want to save files without a trailing newline

src/actions/index.js
```diff
@@ -131,34 +131,20 @@ export function redo() {
...
-
-        // Notify observers
-        dispatch(
-            notifyObservers({
-                id: previous.id,
-                props: previous.props,
-            })
-        );
-    };
+    return undo_revert(UNDO);
```
alexcjohnson: DRYing this section didn't change the functionality at all... but I discovered I could while researching another question: Can we remove the history from the store entirely when these buttons are hidden? The answer to that is no: in principle we could simplify it to saving just the last `past` entry and nothing else - which may be a worthwhile improvement for very large and long-running apps, it's possible the memory usage will eventually cause problems. But `revert` is still needed for error recovery.

Marc-Andre-Rivet: Could you open an issue to investigate/implement the improvement? Wonder how much of an issue this could be today with long-running apps updating graph/table components.

alexcjohnson: > Could you open an issue to investigate/implement the improvement? Wonder how much of an issue this could be today with long-running apps updating graph/table components.

https://github.com/plotly/dash/issues/727

src/AppContainer.react.js
```diff
@@ -30,9 +30,10 @@ class UnconnectedAppContainer extends React.Component {
         if (type(config) === 'Null') {
             return <div className="_dash-loading">Loading...</div>;
         }
+        const {show_undo_redo} = config;
         return (
             <React.Fragment>
-                <Toolbar />
+                {show_undo_redo ? <Toolbar /> : null}
```
Marc-Andre-Rivet: Maybe a note in the Changelog to the effect that that the CSS solution is now obsolete.

tests/test_render.py
```diff
@@ -154,11 +154,12 @@ def test_initial_state(self):
         self.startServer(app)
         el = self.wait_for_element_by_css_selector('#react-entry-point')
 
+        # Note: this .html file shows there's no undo/redo button by default
```
Marc-Andre-Rivet: Due to visual tests instability I would prefer to isolate this behavior in its own test instead of grafting it to an existing one. Specifically, testing that undo & redo buttons are not found in the DOM + a percy snapshot.

alexcjohnson: I'll include a test that these elements aren't in the DOM. The test I already added includes a percy snapshot with undo AND redo buttons - I guess for completeness I could extend it test that at either end of the history the appropriate button disappears, I'll do that - but we now have a million percy snapshots that used to have undo buttons and no longer do.

src/AppContainer.react.js
alexcjohnson: changelog: https://github.com/plotly/dash-renderer/pull/175/commits/e1941e1a468c9abb6d2d12c18c54a19324c6283e

tests/test_render.py
alexcjohnson: extended element existence tests -> https://github.com/plotly/dash-renderer/pull/175/commits/56cfddc62807208eadb4ad18378bd0a37508081c

